PR #160, opened 2015-11-05
Update changelog

Artiom-M:


src/xpiks-qt/debian/changelog
```diff
@@ -1,3 +1,24 @@
+xpiks (1.0-1) UNRELEASED; urgency=medium
```
Ribtoks: Xpiks is released at the moment. So should it be RELEASED?

PR #165, opened 2015-11-19
To be able to disable automatic updates

Artiom-M:
As an alternative it can be an config option.


src/xpiks-qt/main.cpp
```diff
@@ -255,8 +255,9 @@ int main(int argc, char *argv[]) {
     }
 #endif
 
+#ifndef NO_AUTO_UPDATE
```
Ribtoks: Did you forget to patch .pro file with compile-time define as well?

Artiom-M: Will do it for the next release.

------ Original Message ------

Ribtoks:
Will be fixed with Q_OS_LINUX in the other way.

PR #226, opened 2016-01-24
Exit sounds confusing here.

Artiom-M:
User just want to close the Settings dialog. This is what exactly happens.


Ribtoks:
Ok, no problem.

PR #232, opened 2016-01-27
Debian package fixes

Artiom-M:
This time the change list is long. There are new options, lot of 'include' and linker fixes.
.deb works in Ubuntu with the same quirks as previous. Please update the site with the list of dependencies from INSTALL.md
https://mega.nz/#!y5N0HLCL!tNCD9icT09wyaQz5KTt8AYYEvLjnL4jkbrV0WC7R2hs

I still have some troubles with .rpm created by alien. Somehow It breaks my virtual Opensuse...

### DIFF
--- a/src/xpiks-qt/Commands/commandmanager.cpp
+++ b/src/xpiks-qt/Commands/commandmanager.cpp
@@ -176,6 +176,9 @@ void Commands::CommandManager::connectEntitiesSignalsSlots() const {
 
     QObject::connect(m_ArtItemsModel, SIGNAL(selectedArtworkRemoved()),
                      m_FilteredItemsModel, SLOT(onSelectedArtworksRemoved()));
+
+    QObject::connect(m_SettingsModel, SIGNAL(allValuesSaved()),
+                     m_TelemetryService, SLOT(changeReporting()));
 }
 
 void Commands::CommandManager::recodePasswords(const QString &oldMasterPassword,
--- a/src/xpiks-qt/Conectivity/telemetryservice.cpp
+++ b/src/xpiks-qt/Conectivity/telemetryservice.cpp
@@ -31,6 +31,7 @@
 #include <QNetworkReply>
 #include "telemetryservice.h"
 #include "analyticsuserevent.h"
+#include "Models/settingsmodel.h"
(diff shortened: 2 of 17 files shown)


Artiom-M:
And spell worker restart is not implemented. Too complicated for me...

Ribtoks:
Thank you for your work! I had some requests from users who were waiting for the new Linux package and now I'll notify them that it's available!

P.S. Dictionaries directory selection should be visible only to linux users - I'll add appropriate checks there. All other platforms receive dictionaries with Xpiks - no need to select new directory. 
P.P.S. Sure, no problem. I'll implement worker restart.
P.P.P.S. I don't want users to be able to disable simple telemetry in Settings an the moment. Only by recompiling Xpiks.

Ribtoks:
Did you try to run it and play around? I have crashes under opensuse 13.2

Ribtoks:
Please, try to pull latest changes, recompile .deb and then .rpm with alien and try it. I think "exiftool" dependency somehow failed to go to .rpm and I've found a bug that caused a crash (basically, not a bug but my assert).

Ribtoks:
Also please, do not write `not defined()`, but use `!defined()` instead. clang has some problems with it.

Artiom-M:
I didn't test RPM, I have troubles with alien. Did you succeed to convert deb to rpm?

Ribtoks:
No, I just built it via Qt Creator under openSUSE 13.2. I think that you can skip converting .deb to .rpm. It will always fail to correctly translate dependencies. We need normal RPM (Fedora-based or openSUSE-based).

Artiom-M:
I had the same conclusion, so I started to configure a build environment for RPMs. It looks even less complicated than for deb. So I think for the next version there will be an RPM as well.

Ribtoks:
Cool, thank you!

Ribtoks:
Btw, next version will be sooner than you expect. I've implemented a new robust ftp uploader on top of libcurl instead of curl binary and I will merge it back to master this week along with some minor UI fixes, it will go to 1.1.4 bugfix release.

Afterwards, there will be a big pause before the 1.2 release (with the new features).

PR #235, opened 2016-01-29
Debian fixes 1.1.3-2

Artiom-M:
Here is the new package and the changes.
https://mega.nz/#!rl1RhARZ!8KfmUTgfKfenI2t1GVEXfBiRa44BHbP6DfYlZUBrLIk
Future suggestions:
- if you add a new library to include please use _INCLUDEPATH_ variable in .pro file to specify the path to its headers, then in `#include` use just `<name.h>`. This is portable way for headers inclusion.

- disable/enable compile time features from .pro file by adding triggers to _DEFINE_ variable. This avoids useless code modification.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -16,13 +16,13 @@ Exiftool and curl are packed inside the zip archive. Download and extract Xpiks 
 Install the deb file from latest release with all dependencies.
 
 # Special instructions #
+**Linux**
+
+Install all recommended packages:
+- libhunspell
+- hunspell-en-us
 
 **Ubuntu**
-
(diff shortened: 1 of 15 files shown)


Artiom-M:
In Suse section this line:
` LIBS += /usr/lib64/libcurl.so.4` 
has no special effect different from this line: 
`LIBS += -L/usr/lib64/`
If you want to force link against a static library this have to be *.a* file.

Ribtoks:
Regarding SUSE's .a file, it's not supplied with -devel package. So I just put .so there.

Artiom-M:
This to say that adding dynamic libraries to `LIBS` is useless. If there is no error about missing library of cause. By default all libraries are linked dynamically.

PR #240, opened 2016-01-31
iss209

RostaTasha:
Added functionality and tests.

### DIFF
--- a/src/xpiks-qt/Models/filteredartitemsproxymodel.cpp
+++ b/src/xpiks-qt/Models/filteredartitemsproxymodel.cpp
@@ -324,7 +324,7 @@ namespace Models {
         emit allItemsSelectedChanged();
     }
 
-    void FilteredArtItemsProxyModel::reverseFilteredItemsSelected() {
+    void FilteredArtItemsProxyModel::invertFilteredItemsSelected() {
         ArtItemsModel *artItemsModel = getArtItemsModel();
         QVector<int> indices;
         int size = this->rowCount();
--- a/src/xpiks-qt/Models/filteredartitemsproxymodel.cpp
+++ b/src/xpiks-qt/Models/filteredartitemsproxymodel.cpp
@@ -337,10 +337,7 @@ namespace Models {
             int index = originalIndex.row();
             ArtworkMetadata *metadata = artItemsModel->getArtwork(index);
             Q_ASSERT(metadata != NULL);
-            if (metadata->getIsSelected())
-                metadata->setIsSelected(false);
-            else
-                metadata->setIsSelected(true);
+            metadata->invertSelection();
             indices << index;
         }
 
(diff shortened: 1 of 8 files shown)


src/xpiks-tests/filteredmodel_tests.h
```diff
@@ -0,0 +1,21 @@
...
+#include <QtTest/QTest>
+#include "Mocks/artitemsmodelmock.h"
+#include "Mocks/commandmanagermock.h"
+#include "../xpiks-qt/Models/filteredartitemsproxymodel.h"
+
+class FilteredModelTests : public QObject
+{
+    Q_OBJECT
+public:
+    void invertSelectionTest();
```
Ribtoks: Please, declare test methods in section of private slots. Other sections can be dismissed as they are not used.

src/xpiks-qt/Models/filteredartitemsproxymodel.cpp
```diff
@@ -324,6 +324,31 @@ namespace Models {
...
+        indices.reserve(size);
+
+        for (int row = 0; row < size; ++row) {
+            QModelIndex proxyIndex = this->index(row, 0);
+            QModelIndex originalIndex = this->mapToSource(proxyIndex);
+
+            int index = originalIndex.row();
+            ArtworkMetadata *metadata = artItemsModel->getArtwork(index);
+            Q_ASSERT(metadata != NULL);
+            if (metadata->getIsSelected())
```
Ribtoks: Please, try to use code convention where you put brackets even for one-line if statement. 

Also, I can suggest you to edit this line to smth like `metadata->setIsSelected(!metadata->getIsSelected());` because each `if` statement slows down your code. But the ultimate best option would be to write method `invertSelection()` for `ArtworkMetadata`, and let it handle it by itself.

src/xpiks-qt/main.qml
```diff
@@ -240,6 +240,14 @@ ApplicationWindow {
                     openUploadDialog()
                 }
             }
+
+            MenuItem {
+                text: qsTr("&Invert selection")
+                onTriggered: {
+                    console.log("Invert selection triggered")
+                    filteredArtItemsModel.reverseFilteredArtworks()
```
Ribtoks: There's an inconsistency in naming here. MenuItem says "_Invert selection_" but your method name is `reverseFilteredArtworks()`. Also it's not quite correct from English point of view, because reverse changes order of elements to the opposite and `invertSelection` inverts selection

src/xpiks-tests/filteredmodel_tests.cpp
```diff
@@ -0,0 +1,68 @@
+#include "filteredmodel_tests.h"
+
+void FilteredModelTests::invertSelectionTest()
```
Ribtoks: Please, split this method to several methods. Like `invertSelectionForEmptyTest()`, `invertSelectionForSingleTest()`, `invertSelectionForAllSelectedTest()` and etc. Each method (and test method as well) should fit into one screen of code.

RostaTasha:
took into account comments

PR #247, opened 2016-02-09
Update notification service for Linux

Artiom-M:
Now I remember that I wanted this to be an option. So here it is. One thing still missing: custom notification for linux.


Ribtoks:
Yes, I also wanted to have this. Just it wasn't a priority. Thank you for your work.

The only thing which is doubtful is the layout - height of SettingsWindow is growing. Maybe I'll change something.

Artiom-M:
I changed the height of SettingsWindow in order to fit a new option.

Ribtoks:
Yes, you did, but now other tabs will look different with new height. I tried to have at least some of them equally filled. It's not a big deal. Just saying.

PR #251, opened 2016-02-17
231 seems to work

RostaTasha:


Ribtoks:
There are several issues with your code.

1. Why do you need a loop? Log line starts with time which is "dd.MM.yyyy hh:mm:ss.zzz", a dash with spaces " - " and then with log entry severity. You can calculate how many characters can you skip before parsing severity.

2. To compare strings with constants in Qt, please use `if (myString == QLatin1Literal("String Constant"))` because in the other case Qt will each time create a new variable of type QString (it will cast `const char *` to `QString`.

3. Colors. I don't remember, but I must have told you to use colors from `Constants/Colors.js` for Xpiks to have only one set of used colors. So for red you can use "destructiveColor", for orange "artworkModifiedColor", for white just don't do anything (inside it uses "defaultLightColor"), for gray you should choose by yourself either "defaultInputBackground" or "itemsSourceForeground" or "selectedMetadataColor". Also you can create colors in the constructor (as private variables) to avoid instantiation of all colors for each line of text.

4. All classes inside Xpiks are in the namespaces. It can prevent names conflicts in future. Namespace name should correspond to the directory name. Please, put logs highlighter in the namespace "Helpers".

5. There is something wrong with spaces in the `LogHighlighter::highlightBlock()`. Inside while loop is too far to the right and the final brace is outspaced too. While still you are removing while loop, please ensure spaces are fine before committing.

RostaTasha:
Hope I took everything into account.

Ribtoks:
Yes, you did. But there's two more things. First one - private variables should be in format `m_CamelCase`.

And another one. Method/function should have only 1 return statement, because otherwise it's easy to write errors. If we have several consecutive `if` statements in order to make execution optimal, we have to order them in way from the most probable to the least probable. The most common case is to have Debug, then Info, then Warning, then Critical and Fatal. And `if` statements should be in very this order.

Also, it's bad each time to acquire substring with method `mid()`. You're doing same work several times and memory is copied several times. It's better to get the largest possible case (critical) and to compare values using `startsWith()`.
     
    QString logType = word.mid(0,11);
    if (logType.startsWith(QLatin1String("Debug"))) {
    } else if (... startsWith "Info") {
    } ....

RostaTasha:
Made changes according to last comment.

PR #329, opened 2016-08-08
Linux package 1.3.4

Artiom-M:
I will update binary packages as well


Ribtoks:
Why did you remove `property bool useStatistics: settingsModel.userStatistics` from SettingsWindow.qml?
As well as `onCheckedChanged` handler. Please see the diff to this pull request at GitHub

Ribtoks:
And `- LIBS -= -lexiv2` from win32 section. A typo?

Artiom-M:
There was a bug in settings: statistics option status was not saved. I copied a similar section from another tab (that's why it was connected to behavour) and it worked. onCheckedChanged call is there, just order was changed.

win32 section should be a result of the merge with master and conflict resolving... so I will put it back

Ribtoks:
1. I would kindly ask you if there's a bug then create an issue in the Issues section with steps to reproduce and proof it doesn't work (logs where it's clear that after disabling this option, statistics is still being sent) - I will gladly fix that for 1.4

2. I think that there's no bug there. Maybe just some misunderstanding of how it works.

3. In your commit you introduced a bug. What if I check/uncheck the checkbox and quit Settings without saving? Option should not be saved, but it will be with your changes. Using your code it will be saved. I would suggest you to revert your "fix", create an Issue and I will investigate that and resolve it for the 1.4 since now it's arguable.

Ribtoks:
I like your fix for RunGuard for different users. It would be nice to forward it to master too. Can you confirm it will work out of the box in Windows and OS X?

Artiom-M:
I can not confirm it on Windows or OS X. I am almost sure it will
      work with OS X. For windows it will not be worse than now.
    
    
    On 08/08/2016 10:13, Taras Kushnir
      wrote:
    
    
      I like your fix for RunGuard for different users. It would be
        nice to forward it to master too. Can you confirm it will work
        out of the box in Windows and OS X?
      —
        You are receiving this because you authored the thread.
        Reply to this email directly, view
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
      {"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/Ribtoks/xpiks","title":"Ribtoks/xpiks","subtitle":"GitHub repository","main_image_url":"https://assets-cdn.github.com/images/modules/aws/aws-bg.jpg","avatar_image_url":"https://cloud.githubusercontent.com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-b52498112777.png","action":{"name":"Open in GitHub","url":"https://github.com/Ribtoks/xpiks"}},"updates":{"snippets":[{"icon":"PERSON","message":"@Ribtoks in #329: I like your fix for RunGuard for different users. It would be nice to forward it to master too. Can you confirm it will work out of the box in Windows and OS X?"}],"action":{"name":"View Pull Request","url":"https://github.com/Ribtoks/xpiks/pull/329#issuecomment-238169616"}}}

Artiom-M:
3. The bug was not about statistics sent or not sent. I was about the option is not saved and is always checked. I took the same code as you use for other options and it works (at least visually) as expected. The option is saved with "Save" button and discarded with close.

Ribtoks:
Yes, I can see that option is not saved. This is because of the renaming of your _statistic_ to _statistics_. I've pushed a fix to backports_1.3.3 branch.

Why your fix introduces a bug:
in my code, checkbox values are saved when Save button is pressed. In your code, checkbox value is saved immediately. So steps to reproduce a bug using your code: (let's suppose statistics is ON)

- Open Settings
- uncheck the statistics checkbox
- press "Close" (without saving)

Expected result: No changes are saved.
Real result: you have saved the statistics option.

That's why you should revert your fix (anyway, I've pushed a typo fix)

Artiom-M:
I uploaded packages built on this branch.

Ribtoks:
Thank you

Issue #290, opened 2016-05-05
Add "misspelled" word to dictionary


Ribtoks:
Create and maintain user-defined dictionary for words so that spellchecking will not mark them as wrong.

Word can be added to the dictionary through the context menu available only for misspelled words.

PR #333, opened 2016-08-17
relates #290

RostaTasha:


Ribtoks:
Adding wrong word from Description/Title is not yet implemented.

PR #348, opened 2016-10-01
closes #337

### LINKED ISSUE #337: Create server-side config for some warnings parameters
Create parameters for min/max size of images and others

RostaTasha:


RostaTasha:
Example of json config file
`{
    "overwrite": false,
    "settings": {
        "chars": "abc",
        "max_description_length": 1,
        "max_keywords_count": 2,
        "min_mpixels": 0
    },
    "version": 2
}`

src/xpiks-qt/AutoComplete/warningssettingsmodel.cpp
```diff
@@ -0,0 +1,142 @@
...
+            m_MinMegapixels = minMPixels.toDouble();
+
+            QJsonValue maxKeywordsCount = settingsObject[MAX_KEYWORDS_COUNT];
+            if (!maxKeywordsCount.isDouble()) {
+                LOG_WARNING << "MAX_KEYWORDS_COUNT value is not number";
+                anyError = true;
+                break;
+            }
+
+            m_MaxKeywordsCount = maxKeywordsCount.toDouble();
```
Ribtoks: Why keywords count is double? Isn't it supposed to be an integer?

src/xpiks-qt/AutoComplete/warningssettingsmodel.cpp
```diff
@@ -0,0 +1,142 @@
...
+            if (!maxDescriptionCount.isDouble()) {
+                LOG_WARNING << "MAX_DESCRIPTION_LENGTH value is not number";
+                anyError = true;
+                break;
+            }
+
+            m_MaxDescriptionLength = maxDescriptionCount.toDouble();
+        } while (false);
+
+        LOG_DEBUG<<anyError;
```
Ribtoks: spaces

src/xpiks-qt/AutoComplete/warningssettingsmodel.cpp
```diff
@@ -0,0 +1,142 @@
...
+            m_MaxKeywordsCount = maxKeywordsCount.toDouble();
+
+            QJsonValue maxDescriptionCount = settingsObject[MAX_DESCRIPTION_LENGTH];
+            if (!maxDescriptionCount.isDouble()) {
+                LOG_WARNING << "MAX_DESCRIPTION_LENGTH value is not number";
+                anyError = true;
+                break;
+            }
+
+            m_MaxDescriptionLength = maxDescriptionCount.toDouble();
```
Ribtoks: same here - length of description is supposed to be an integer value

src/xpiks-qt/Warnings/warningscheckingworker.cpp
```diff
@@ -67,35 +58,39 @@ namespace Warnings {
...
-        m_MaximumDescriptionLength = m_SettingsModel->getMaxDescriptionLength();
-        m_MinimumMegapixels = m_SettingsModel->getMinMegapixelCount();
-        m_MaximumKeywordsCount = m_SettingsModel->getMaxKeywordsCount();
-    }
-
     Common::WarningFlags WarningsCheckingWorker::checkDimensions(std::shared_ptr<WarningsItem> &wi) const {
+        QString allowedFilenameCharacters;
+        double minimumMegapixels = 0;
+
+        if (m_WarningsSettingsModel != nullptr) {
```
Ribtoks: It will be good not to check each time that m_WarningsSettingsModel is not NULL. Programmer is responsible for passing not null so you can only use Q_ASSERT.

src/xpiks-qt/Warnings/warningscheckingworker.cpp
```diff
@@ -39,15 +38,9 @@ namespace Warnings {
...
+    WarningsCheckingWorker::WarningsCheckingWorker(AutoComplete::WarningsSettingsModel *warningsSettingsModel, QObject *parent):
         QObject(parent),
-        m_SettingsModel(settingsModel),
-        m_AllowedFilenameCharacters(".,_-@ "),
-        m_MinimumMegapixels(4),
-        m_MaximumKeywordsCount(50),
-        m_MaximumDescriptionLength(200)
-    {
-    }
+        m_WarningsSettingsModel(warningsSettingsModel) {}
```
Ribtoks: This should be the entry point for checking if WarningsSettingsModel is not null - I mean Q_ASSERT

src/xpiks-qt/Models/settingsmodel.h
```diff
@@ -372,11 +339,8 @@ namespace Models {
         QString m_ExifToolPath;
         QString m_DictPath;
         QString m_SelectedLocale;
-        double m_MinMegapixelCount;
         double m_KeywordSizeScale;
         double m_ScrollSpeedScale;
-        int m_MaxDescriptionLength;
```
Ribtoks: If you have removed everything Warnings-related from the model, then you will need to remove Warnings Tab from SettingsWindow as well.

src/xpiks-qt/AutoComplete/warningssettingsmodel.cpp
```diff
@@ -0,0 +1,142 @@
+#include "warningssettingsmodel.h"
```
Ribtoks: Please add GPL 3.0 header like in the other files.

src/xpiks-qt/AutoComplete/warningssettingsmodel.h
```diff
@@ -0,0 +1,41 @@
+#ifndef WARNINGSSETTINGSMODEL_H
```
Ribtoks: Please add GPL 3.0 header like in the other files.

Ribtoks:
And please also initiate remote config update 1 second after the launch instead of immediately. You can achieve that using `QTimer::singleShot(1000, this, SLOT(updateStocksList()));` in the file _artworkuploader.cpp_

PR #355, opened 2016-12-19
207

RostaTasha:
without tests yet. Just to make sure, that the implementation looks ok.


src/xpiks-qt/Models/filteredartitemsproxymodel.cpp
```diff
@@ -386,6 +386,65 @@ namespace Models {
         return result;
     }
 
+    void FilteredArtItemsProxyModel::appendFromPreset(int index, int presetIndex)
+    {
+        if (0 <= index && index < rowCount()) {
+            int originalIndex = getOriginalIndex(index);
+            LOG_INFO << originalIndex;
```
Ribtoks: it's good to log in the root scope also the first `index` and `presetindex` so we will know the call was done to this method and if the other part did not execute, we will know which index was requested which did not correspond to `originalIndex`

src/xpiks-qt/Models/filteredartitemsproxymodel.cpp
```diff
@@ -386,6 +386,65 @@ namespace Models {
...
+    }
+
+    void FilteredArtItemsProxyModel::replaceFromPreset(int index, const QString & keywordsName, int presetIndex)
+    {
+        if (0 <= index && index < rowCount()) {
+            int originalIndex = getOriginalIndex(index);
+            LOG_INFO << originalIndex;
+            auto presetModel = m_CommandManager->getPresetModel();
+            QString presetName = presetModel->getNameFromIndex(presetIndex);
+            if (presetName == QString()){
```
Ribtoks: Is this some creative sort of comparing to empty string? If yes, this one reads a bit worse than `string.isEmpty()`

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,230 @@
...
+    PresetKeywordsModel::PresetKeywordsModel(QObject *parent):
+        QAbstractListModel(parent),
+        Common::BaseEntity()
+    {}
+
+    int PresetKeywordsModel::getIndexFromName(const QString &presetName)
+    {
+        int result = -1;
+        int size = m_Presets.size();
+        for (int i =0; i < size; i++){
```
Ribtoks: please format this function better (spaces, ampersands, opening brace)

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+    QString PresetKeywordsModel::getNameFromIndex(int index)
+    {
+        if (index < 0 || index >= m_Presets.size()){
+            return QString();
+        }
+        return m_Presets[index].name;
+    }
+
+    QStringList PresetKeywordsModel::getKeywords(int index)
```
Ribtoks: Is it this API which can be improved with "try" thing I've proposed earlier?

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,230 @@
...
+        beginRemoveRows(QModelIndex(), row, row);
+        removeInnerItem(row);
+        endRemoveRows();
+    }
+
+    void PresetKeywordsModel::addItem() {
+        int lastIndex = (int)getPresetsCount();
+        std::shared_ptr<Common::BasicKeywordsModel> ptr(new Common::BasicKeywordsModel(m_HoldPlaceholder));
+        beginInsertRows(QModelIndex(), lastIndex, lastIndex);
+        m_Presets.push_back({{},QString("Undefined")});
```
Ribtoks: std::vector has `emplace_back()` which invokes move semantics. This code (and line below) is not very good for reading.

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,230 @@
...
+        m_Presets.last().keys.swap(ptr);
+        endInsertRows();
+        LOG_INFO << "";
+        emit dataChanged(this->index(m_Presets.size()-1),this->index(m_Presets.size()-1));
```
Ribtoks: You have `lastIndex` cached earlier - you can use it for parameters - no need to write .size() twice)

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,230 @@
...
+        if (0 <= index && index < getPresetsCount()) {
+            auto & keywordsModel = m_Presets[index].keys;
+            if (keywordsModel->takeKeywordAt(keywordIndex, keyword)) {
+                LOG_INFO << "Removed keyword:" << keyword;
+            }
+        }
+     return keyword;
+    }
+
+    void PresetKeywordsModel::plainTextEdit(int index, const QString &rawKeywords, bool spaceSeparator) {
```
Ribtoks: Let's not allow users to edit presets in plain text. Keywords editing is pretty good and let's keep the UI simple. What do you think?

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,230 @@
...
+        presetConfig->saveFromModel(m_Presets);
+       #endif
+    }
+
+    void PresetKeywordsModel::resetModel()
+    {
+        auto presetConfig = m_CommandManager->getPresetModelConfig();
+        auto & presetData = presetConfig->m_PresetData;
+        int size = presetData.size();
+        LOG_WARNING<<"reset preset Model "<<size;
```
Ribtoks: Warning level is usually when something is wrong. For logging with parameters please use "info"

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,230 @@
...
+                presets.push_back(name);
+            }
+        }
+
+        return presets;
+    }
+
+    int PresetKeywordsModel::rowCount(const QModelIndex &parent) const {
+        Q_UNUSED(parent);
+        LOG_DEBUG << "tasha ok" << m_Presets.size();
```
Ribtoks: Fancy logging text

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.h
```diff
@@ -0,0 +1,58 @@
...
+#include "../Common/basickeywordsmodel.h"
+#include "../Common/baseentity.h"
+#include "../Common/abstractlistmodel.h"
+#include <QAbstractListModel>
+#include <QVector>
+
+namespace Preset {
+
+struct Preset {
+   std::shared_ptr<Common::BasicKeywordsModel> keys;
```
Ribtoks: Can you please follow naming conventions like `m_KeywordsModel` and `m_PresetName` instead of `keys` and `name`

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.h
```diff
@@ -0,0 +1,58 @@
...
+        Q_INVOKABLE void editKeyword(int index, int keywordIndex, const QString &replacement);
+        Q_INVOKABLE QString removeKeywordAt(int index, int keywordIndex);
+        Q_INVOKABLE void plainTextEdit(int index, const QString &rawKeywords, bool spaceSeparator);
+        Q_INVOKABLE void appendKeyword(int index, const QString &keyword);
+        Q_INVOKABLE QObject *getKeywordsModel(int index);
+        Q_INVOKABLE void saveToConfig();
+        Q_INVOKABLE void resetModel();
+        Q_INVOKABLE QStringList getFilteredPresets(const QString &  word);
+    public slots:
+       void presetsUpdated() {LOG_WARNING<<"preset Model"; resetModel();}
```
Ribtoks: I'm not sure naming conventions work here. For example, I don't follow the logic that when the presets are updated we need to reset model.

src/xpiks-qt/PresetKeywords/presetkeywordsmodelconfig.cpp
```diff
@@ -0,0 +1,154 @@
...
+            auto keywords = keywordsModel->getKeywords();
+            m_PresetData[i].keys = keywords;
+            m_PresetData[i].name = name;
+        }
+
+        writeToConfig();
+    }
+
+    bool PresetKeywordsModelConfig::parseConfig(const QJsonDocument &document) {
+#ifdef QT_DEBUG
```
Ribtoks: There's appropriate macro LOG_INTEGR_TESTS_OR_DEBUG for that. Not quite readable, but works.

src/xpiks-qt/PresetKeywords/presetkeywordsmodelconfig.cpp
```diff
@@ -0,0 +1,154 @@
...
+            if (!presetKeywordsValue.isArray()) {
+                continue;
+            }
+
+            QJsonArray jsonArray = presetKeywordsValue.toArray();
+            QStringList list;
+            int size = jsonArray.size();
+            list.reserve(size);
+            for (auto &item: jsonArray) {
+                if (!item.isString()) {
```
Ribtoks: We should log this with warning level, I suppose. We handle the config and if the structure is not like we have left it - than something is wrong - user has opened in or smth.

src/xpiks-qt/PresetKeywords/presetkeywordsmodelconfig.cpp
```diff
@@ -0,0 +1,154 @@
...
+                list.append(item.toString());
+            }
+            m_PresetData.push_back({list, presetKey});
+        }
+    }
+
+    void PresetKeywordsModelConfig::writeToConfig()
+    {
+      QJsonArray jsonArray;
+      for (auto & item : m_PresetData){
```
Ribtoks: Please fix the spacing

src/xpiks-qt/main.cpp
```diff
@@ -86,7 +86,8 @@
 #include "MetadataIO/exiv2inithelper.h"
 #include "Models/findandreplacemodel.h"
 #include "Models/previewmetadataelement.h"
-
+#include "PresetKeywords/presetkeywordsmodel.h"
+#include "PresetKeywords/presetkeywordsmodelconfig.h"
 void myMessageHandler(QtMsgType type, const QMessageLogContext &context, const QString &msg) {
```
Ribtoks: Please leave some space before the function

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -487,6 +498,24 @@ namespace Common {
         return anythingRemoved;
     }
 
+    bool BasicKeywordsModel::replaceKeywordsUnsafe(const QString &replaceFrom, const QStringList &keywordsList) {
+        int size = m_KeywordsList.size();
+        bool result = false;
+        LOG_DEBUG << "tashatasha " << replaceFrom << " " << keywordsList;
+
+        for (int i = 0; i < size; ++i) {
+            auto & internal = m_KeywordsList.at(i);
```
RostaTasha: Done

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -487,6 +498,24 @@ namespace Common {
         return anythingRemoved;
     }
 
+    bool BasicKeywordsModel::replaceKeywordsUnsafe(const QString &replaceFrom, const QStringList &keywordsList) {
+        int size = m_KeywordsList.size();
+        bool result = false;
+        LOG_DEBUG << "tashatasha " << replaceFrom << " " << keywordsList;
```
RostaTasha: Done

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -173,6 +173,17 @@ namespace Common {
         setKeywordsUnsafe(keywordsList);
     }
 
+    bool BasicKeywordsModel::replaceKeywords(const QString &replaceFrom, const QStringList &keywordsList)
```
RostaTasha: Done

src/xpiks-qt/Commands/commandmanager.h
```diff
@@ -311,6 +319,8 @@ namespace Commands {
         Models::DeleteKeywordsViewModel *m_DeleteKeywordsViewModel;
         Models::FindAndReplaceModel *m_FindAndReplaceModel;
         Helpers::HelpersQmlWrapper *m_HelpersQmlWrapper;
+        Preset::PresetKeywordsModel *m_PresetModel;
```
RostaTasha: Done

src/xpiks-qt/Commands/commandmanager.cpp
```diff
@@ -281,6 +293,10 @@ void Commands::CommandManager::connectEntitiesSignalsSlots() const {
         QObject::connect(m_UpdateService, SIGNAL(updateDownloaded(QString)),
                          m_HelpersQmlWrapper, SLOT(updateIsDownloaded(QString)));
     }
+
+    if (m_PresetModel != NULL && m_PresetModelConfig != NULL) {
+        QObject::connect(m_PresetModelConfig, SIGNAL(presetsUpdated()), m_PresetModel, SLOT (presetsUpdated()));
```
RostaTasha: Done

src/xpiks-qt/Commands/commandmanager.cpp
```diff
@@ -197,6 +197,18 @@ void Commands::CommandManager::InjectDependency(Helpers::HelpersQmlWrapper *help
     m_HelpersQmlWrapper->setCommandManager(this);
 }
 
+void Commands::CommandManager::InjectDependency(Preset::PresetKeywordsModel *presetModel)
+{
+    Q_ASSERT(presetModel != NULL); m_PresetModel = presetModel;
```
RostaTasha: Done

src/xpiks-qt/Models/combinedartworksmodel.cpp
```diff
@@ -305,6 +305,45 @@ namespace Models {
         return m_CommonKeywordsModel.hasDescriptionWordSpellError(word);
     }
 
+    void CombinedArtworksModel::replaceFromPreset(const QString &keywordsName, int presetIndex)
+    {
+        auto presetModel = m_CommandManager->getPresetModel();
```
RostaTasha: Done

src/xpiks-qt/Models/abstractconfigupdatermodel.cpp
```diff
@@ -39,8 +39,9 @@ namespace Models {
         Q_ASSERT(m_CommandManager != NULL);
         Models::SettingsModel *settingsModel = m_CommandManager->getSettingsModel();
         ProxySettings *proxySettings = settingsModel->retrieveProxySettings();
-
-        m_RemoteConfig.requestInitConfig(configUrl, proxySettings);
+        if (!configUrl.isEmpty()){
```
RostaTasha: Done

src/xpiks-qt/Models/combinedartworksmodel.cpp
```diff
@@ -305,6 +305,45 @@ namespace Models {
         return m_CommonKeywordsModel.hasDescriptionWordSpellError(word);
     }
 
+    void CombinedArtworksModel::replaceFromPreset(const QString &keywordsName, int presetIndex)
+    {
+        auto presetModel = m_CommandManager->getPresetModel();
+        QStringList keywords = presetModel->getKeywords(presetIndex);
```
RostaTasha: Done

src/xpiks-qt/Models/combinedartworksmodel.h
```diff
@@ -179,6 +179,9 @@ namespace Models {
 
         Q_INVOKABLE bool hasTitleWordSpellError(const QString &word);
         Q_INVOKABLE bool hasDescriptionWordSpellError(const QString &word);
+        Q_INVOKABLE void replaceFromPreset(const QString & keywordsName, int presetIndex);
```
RostaTasha: Done

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -487,6 +498,24 @@ namespace Common {
...
+        int size = m_KeywordsList.size();
+        bool result = false;
+        LOG_DEBUG << "tashatasha " << replaceFrom << " " << keywordsList;
+
+        for (int i = 0; i < size; ++i) {
+            auto & internal = m_KeywordsList.at(i);
+            if (internal != replaceFrom) {
+                continue;
+            }
+            removeKeywordsAtIndicesUnsafe({i});
```
RostaTasha: Done

src/xpiks-qt/main.cpp
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodelconfig.cpp
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodelconfig.cpp
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodelconfig.cpp
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.h
RostaTasha: Fixed

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.h
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/filteredartitemsproxymodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/filteredartitemsproxymodel.cpp
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
RostaTasha: QVector does not have emplace_back(). Changed to move operator

src/xpiks-qt/Models/combinedartworksmodel.cpp
```diff
@@ -305,6 +305,46 @@ namespace Models {
         return m_CommonKeywordsModel.hasDescriptionWordSpellError(word);
     }
 
+    void CombinedArtworksModel::replaceFromPreset(const QString &keywordsName, int presetIndex)
+    {
+        auto *presetModel = m_CommandManager->getPresetModel();
+        QStringList keywords;
+        if (!presetModel->tryGetPreset(presetIndex, keywords)){
```
Ribtoks: This is better, but it usually is written like this:

    if (tryGetValue(out value)) {
        // do something with value
    }

It's better not to use `return` s

src/xpiks-qt/Models/combinedartworksmodel.cpp
```diff
@@ -305,6 +305,46 @@ namespace Models {
...
+        m_CommonKeywordsModel.changeFromPreset(keywordsName, keywords);
+        emit keywordsCountChanged();
+        setKeywordsModified(true);
+        m_CommandManager->submitItemForSpellCheck(&m_CommonKeywordsModel, Common::SpellCheckFlags::Keywords);
+    }
+
+    void CombinedArtworksModel::replaceFromPreset(const QString &keywordsName, const QString &presetName)
+    {
+        auto *presetModel = m_CommandManager->getPresetModel();
+        int presetIndex = presetModel->getIndexFromName(presetName);
```
Ribtoks: I guess you can move this to the same `tryGet` or `tryFind` API as previous one

src/xpiks-qt/Models/combinedartworksmodel.cpp
```diff
@@ -305,6 +305,46 @@ namespace Models {
...
+        QStringList keywords;
+        presetModel->tryGetPreset(presetIndex, keywords);
+        m_CommonKeywordsModel.changeFromPreset(keywordsName, keywords);
+        emit keywordsCountChanged();
+        setKeywordsModified(true);
+        m_CommandManager->submitItemForSpellCheck(&m_CommonKeywordsModel, Common::SpellCheckFlags::Keywords);
+    }
+
+    void CombinedArtworksModel::appendFromPreset(int presetIndex) {
+        auto presetModel = m_CommandManager->getPresetModel();
```
Ribtoks: Can you please change this to `auto *presetModel` so it will be clear that `getPresetModel()` returns a pointer?

src/xpiks-qt/Commands/commandmanager.h
```diff
@@ -272,6 +278,8 @@ namespace Commands {
         virtual Models::ZipArchiver *getZipArchiver() const { return m_ZipArchiver; }
         virtual Models::DeleteKeywordsViewModel *getDeleteKeywordsModel() const { return m_DeleteKeywordsViewModel; }
         virtual SpellCheck::SpellCheckSuggestionModel *getSpellSuggestionsModel() const { return m_SpellCheckSuggestionModel; }
+        virtual Presets::PresetKeywordsModel *getPresetModel() const { return m_PresetsModel; }
```
Ribtoks: Same story:
get _Preset_ Model returns m_ _PresetS_ Model. Please change to `getPresetsModel()`

src/xpiks-qt/Models/filteredartitemsproxymodel.cpp
```diff
@@ -386,6 +386,74 @@ namespace Models {
         return result;
     }
 
+    void FilteredArtItemsProxyModel::appendFromPreset(int index, int presetIndex)
+    {
+        if (0 <= index && index < rowCount()) {
+            int originalIndex = getOriginalIndex(index);
+            LOG_INFO << "preset index " << presetIndex << "original index" << originalIndex;
```
Ribtoks: No neeed to use ending with spaces in logging: they're added automatically "preset index " -> "preset index"

src/xpiks-qt/Models/filteredartitemsproxymodel.cpp
```diff
@@ -386,6 +386,74 @@ namespace Models {
...
+            auto presetModel = m_CommandManager->getPresetModel();
+            QStringList keywords;
+            if (!presetModel->tryGetPreset(presetIndex, keywords)){
```
Ribtoks: Same stuff with `return`s

src/xpiks-qt/Models/filteredartitemsproxymodel.cpp
```diff
@@ -386,6 +386,74 @@ namespace Models {
...
+            ArtworkMetadata *metadata = artItemsModel->getArtwork(originalIndex);
+
+            if (metadata != NULL) {
+                auto *keywordsModel = metadata->getBasicModel();
+                keywordsModel->changeFromPreset(keywords);
+            }
+        }
+    }
+
+    void FilteredArtItemsProxyModel::replaceFromPreset(int index, const QString & keywordsName, int presetIndex)
```
Ribtoks: Please follow the braces + ampersand style

src/xpiks-qt/Models/filteredartitemsproxymodel.cpp
```diff
@@ -386,6 +386,74 @@ namespace Models {
...
+    {
+        if (0 <= index && index < rowCount()) {
+            int originalIndex = getOriginalIndex(index);
+            LOG_INFO << originalIndex;
+            auto presetModel = m_CommandManager->getPresetModel();
```
Ribtoks: `auto *presetModel`

src/xpiks-qt/Models/filteredartitemsproxymodel.cpp
```diff
@@ -386,6 +386,74 @@ namespace Models {
...
+            QString presetName = presetModel->getNameFromIndex(presetIndex);
+            if (presetName.isEmpty()){
```
Ribtoks: This function contains too many returns.. Not necessarily, but this can mean something is architecturaly wrong. I have found it this time: you don't know preset index or some sort of ID (except of name which projects to the index). When using Instantiator in menu, you can easily fill in to some new MenuItem property an index of the preset.

    MenuItem {
        property int presetIndex: index

(each delegate has an index)
This should not only make your life easier, but also simplify and make more solid your APIs

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,213 @@
...
+                result = i;
+                break;
+            }
+        }
+
+        return result;
+    }
+
+    QString PresetKeywordsModel::getNameFromIndex(int index)
+    {
```
Ribtoks: returning empty string implies checking for emptiness when you call this function. It's better to use `tryGet`-style API which immediately returns `bool`

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,213 @@
...
+        endRemoveRows();
+    }
+
+    void PresetKeywordsModel::addItem() {
+        int lastIndex = (int)getPresetsCount();
+        std::shared_ptr<Common::BasicKeywordsModel> ptr(new Common::BasicKeywordsModel(m_HoldPlaceholder));
+        beginInsertRows(QModelIndex(), lastIndex, lastIndex);
+        m_Presets.push_back({std::move(ptr),QString("Undefined")});
+        endInsertRows();
+        LOG_INFO << "";
```
Ribtoks: Please use info logging for logging with parameters. In usual cases it's useful just to `LOG_DEBUG << "#";`

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,213 @@
...
+    QString PresetKeywordsModel::removeKeywordAt(int index, int keywordIndex)
+    {
+        QString keyword;
+        if (0 <= index && index < getPresetsCount()) {
+            auto & keywordsModel = m_Presets[index].m_KeywordsModel;
+            if (keywordsModel->takeKeywordAt(keywordIndex, keyword)) {
+                LOG_INFO << "Removed keyword:" << keyword;
+            }
+        }
+     return keyword;
```
Ribtoks: Please fix the formatting + braces style in all of the functions in this file

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,213 @@
...
+       #ifndef CORE_TESTS
+        presetConfig->saveFromModel(m_Presets);
+       #endif
+    }
+
+    void PresetKeywordsModel::loadFromConfigModel()
+    {
+        beginResetModel();
+        auto presetConfig = m_CommandManager->getPresetModelConfig();
+        auto & presetData = presetConfig->m_PresetData;
```
Ribtoks: Not enough newlines in this function to read blocks + displaced ampersands + no spacing before braces

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,213 @@
...
+                 m_Presets[i].m_KeywordsModel.swap(ptr);
+                 m_Presets[i].m_KeywordsModel->appendKeywords(keywords);
+            }
+            m_Presets[i].m_PresetName = name;
+        }
+        endResetModel();
+    }
+
+    QStringList PresetKeywordsModel::getFilteredPresets(const QString &word) {
+        QVector<int> indexes;
```
Ribtoks: What are you using indexes for in this function?

src/xpiks-qt/Dialogs/CombinedArtworksDialog.qml
```diff
@@ -83,6 +83,67 @@ CloseRequested")
...
+
+                }
+            }
+        }
+    }
+
+
+    Menu {
+        id: presetsMenu
+        property int maxSize : 2
```
Ribtoks: Seeing presets in submenu I guess it would be nice to show at least 5 presets (if there would be 5). Or even 10 if there would be 10 presets which contain a keyword substring. I think it's not the case in 99.9% of usage

src/xpiks-qt/Common/basicmetadatamodel.h
```diff
@@ -79,6 +79,8 @@ namespace Common {
         bool hasDescriptionWordSpellError(const QString &word);
         bool hasTitleWordSpellError(const QString &word);
         virtual bool hasSpellErrors();
+        void changeFromPreset(const QString &word, const QStringList & presetList);
+        void changeFromPreset(const QStringList & presetList);
```
RostaTasha: Done

src/xpiks-qt/Dialogs/CombinedArtworksDialog.qml
RostaTasha: Done

src/xpiks-qt/Dialogs/EditArtworkHorizontalDialog.qml
```diff
@@ -81,6 +81,69 @@ Item {
...
+            visible: wordRightClickMenu.showAddToDict
+            text: i18.n + qsTr("Add to dictionary")
+            onTriggered: spellCheckService.addWordToUserDictionary(wordRightClickMenu.word);
+        }
+
+
+        Menu{
+            id : presetSubMenu
+            visible: wordRightClickMenu.showExpandPreset
+            title: "Expand as Presets"
```
RostaTasha: Done

src/xpiks-qt/Commands/commandmanager.h
RostaTasha: Done

src/xpiks-qt/Models/combinedartworksmodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/filteredartitemsproxymodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/filteredartitemsproxymodel.cpp
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
RostaTasha: Removed

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/filteredartitemsproxymodel.cpp
RostaTasha: I don't have indexes in cases, when I find presets, which contain this keyword in name. I decided that  implementation of some sort of filtered model would be shooting with a gun into sparrows, so I simply return list of strings (presetNames)

src/xpiks-qt/Models/combinedartworksmodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/combinedartworksmodel.cpp
RostaTasha: Done

src/xpiks-qt/Dialogs/CombinedArtworksDialog.qml
```diff
@@ -83,6 +83,67 @@ CloseRequested")
...
+                id : presetsInstantiator
+                model: wordRightClickMenu.presets
+                onObjectAdded: presetSubMenu.insertItem( index, object )
+                onObjectRemoved: presetSubMenu.removeItem( object )
+                delegate: MenuItem {
+                    text: i18.n + qsTr("\"%1\"").arg(modelData)
+                    onTriggered: {
+                        combinedArtworks.replaceFromPreset(wordRightClickMenu.word, modelData);
+                    }
+
```
RostaTasha: Removed

src/xpiks-qt/Components/EditableTags.qml
```diff
@@ -164,9 +165,15 @@ Flickable {
         anchors.left: parent.left
         anchors.top: parent.top
         width: parent.width
+        acceptedButtons: Qt.LeftButton | Qt.RightButton
         height: flowListView.contentHeight > flowListView.height ? flowListView.contentHeight : flowListView.height
 
         onClicked: {
+            if (mouse.button == Qt.RightButton) {
```
RostaTasha: Done

src/xpiks-qt/Common/basicmetadatamodel.cpp
```diff
@@ -257,6 +257,16 @@ namespace Common {
         return result;
     }
 
+    void BasicMetadataModel::changeFromPreset(const QString &word, const QStringList & presetList)
+    {
```
RostaTasha: Done

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -487,6 +498,26 @@ namespace Common {
         return anythingRemoved;
     }
 
+    bool BasicKeywordsModel::expandPresetUnsafe(const QString &expandFrom, const QStringList &keywordsList) {
```
RostaTasha: See comment below. Caused by the fact, that I did not implement filtered class for presetsModel. When user clicks on expand Preset for this particular keyword, I know only name of preset

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -173,6 +173,17 @@ namespace Common {
         setKeywordsUnsafe(keywordsList);
     }
 
+    bool BasicKeywordsModel::expandPreset(const QString &expandFrom, const QStringList &keywordsList)
+    {
```
RostaTasha: Done

src/xpiks-qt/Commands/commandmanager.cpp
```diff
@@ -197,6 +197,17 @@ void Commands::CommandManager::InjectDependency(Helpers::HelpersQmlWrapper *help
     m_HelpersQmlWrapper->setCommandManager(this);
 }
 
+void Commands::CommandManager::InjectDependency(Presets::PresetKeywordsModel *presetModel) {
+    Q_ASSERT(presetModel != NULL); m_PresetsModel = presetModel;
+    m_PresetsModel->setCommandManager(this);
+}
+
+void Commands::CommandManager::InjectDependency(Presets::PresetKeywordsModelConfig *presetModelConfig)
+{
```
RostaTasha: Done

src/xpiks-qt/Common/basicmetadatamodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/filteredartitemsproxymodel.cpp
RostaTasha: Added filteredModel

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -487,6 +498,26 @@ namespace Common {
         return anythingRemoved;
     }
 
+    bool BasicKeywordsModel::expandPresetUnsafe(const QString &expandFrom, const QStringList &keywordsList) {
+        int size = m_KeywordsList.size();
+        bool result = false;
+        LOG_DEBUG << "target word " << expandFrom << " " << keywordsList;
+
+        for (int i = 0; i < size; ++i) {
```
RostaTasha: Changed

src/xpiks-qt/Common/basickeywordsmodel.cpp
RostaTasha: Fixed

src/xpiks-qt/Common/basicmetadatamodel.cpp
```diff
@@ -257,6 +257,14 @@ namespace Common {
         return result;
     }
 
+    void BasicMetadataModel::changeFromPreset(int keywordsIndex, const QStringList &presetList) {
+        expandPreset(keywordsIndex, presetList);
+    }
+
+    void BasicMetadataModel::changeFromPreset(const QStringList &presetList) {
```
Ribtoks: I would say this name is confusing. This API does not change anything. I believe something like `addFromPreset()` will work a lot better

src/xpiks-qt/Components/EditableTags.qml
```diff
@@ -64,6 +64,7 @@ Flickable {
     signal tabPressed();
     signal editActivated();
     signal clickedInside();
+    signal clickedInsideRight();
```
Ribtoks: Why not `rightClickedInside()`? I would say it makes more sense

src/xpiks-qt/Dialogs/CombinedArtworksDialog.qml
```diff
@@ -83,6 +83,67 @@ CloseRequested")
...
+        MenuItem {
+            visible: wordRightClickMenu.showAddToDict
+            text: i18.n + qsTr("Add to dictionary")
+            onTriggered: spellCheckService.addWordToUserDictionary(wordRightClickMenu.word);
+        }
+
+        Menu {
+            id : presetSubMenu
+            visible: wordRightClickMenu.showExpandPreset
+            title: i18.n + qsTr("Expand as one Preset")
```
Ribtoks: Why do you stress that it is "one" preset? Do you have any arguments against just "Expand as preset"?

src/xpiks-qt/Dialogs/CombinedArtworksDialog.qml
```diff
@@ -83,6 +83,67 @@ CloseRequested")
...
+                    onTriggered: {
+                        combinedArtworks.replaceFromPreset(wordRightClickMenu.keywordIndex, filteredPresetsModel.getOriginalIndex(index));
+                    }
+                }
+            }
+        }
+    }
+
+
+    Menu {
```
Ribtoks: I guess this menu can be removed since there gonna be no "More" link in this release (as we agreed in chat).

src/xpiks-qt/Dialogs/EditArtworkHorizontalDialog.qml
```diff
@@ -81,6 +81,68 @@ Item {
...
+                    onTriggered: {
+                        combinedArtworks.replaceFromPreset(wordRightClickMenu.keywordIndex, filteredPresetsModel.getOriginalIndex(index));
+                    }
+                }
+            }
+        }
+    }
+
+
+    Menu {
```
Ribtoks: Same as above - no "More" in 1.4 release

src/xpiks-qt/Models/combinedartworksmodel.cpp
```diff
@@ -305,6 +305,31 @@ namespace Models {
         return m_CommonKeywordsModel.hasDescriptionWordSpellError(word);
     }
 
+    void CombinedArtworksModel::replaceFromPreset(int keywordsIndex, int presetIndex)
+    {
+        auto *presetsModel = m_CommandManager->getPresetsModel();
+        QStringList keywords;
+        if (!presetsModel->tryGetPreset(presetIndex, keywords)){
```
Ribtoks: Can you please do like you did in `CombinedArtworksModel::appendFromPreset` just below? I mean moving block inside `tryGet` scope instead of returning.

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+    bool PresetKeywordsModel::tryGetIndexFromName(const QString &presetName, int &result) {
+        result = -1;
+        int size = m_Presets.size();
+
+        for (int i = 0; i < size; i++) {
+            auto &item = m_Presets[i];
+            if (item.m_PresetName == presetName) {
+                result = i;
+                return true;
```
Ribtoks: Please try to use "single return" policy as much as possible. For example, here it's very easy to introduce `bool found = false;` and `return found;` in the end only assigning `found = true;` in case of match. This really helps to follow logic instead of looking what could have happened after the loop in case if there was no match and no return.

This function does not crucially need this, but you should better do that in all such places just to do that automatically in complex situations later.

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,255 @@
...
+        if (presetIndex < 0 || presetIndex >= m_Presets.size()) {
+            return false;
+        }
+
+        keywords = m_Presets[presetIndex].m_KeywordsModel->getKeywords();
+        return true;
+    }
+
+    void PresetKeywordsModel::removeItem(int row) {
+        beginRemoveRows(QModelIndex(), row, row);
```
Ribtoks: This is Q_IVOKABLE, public method. You should always validate input parameters of public API and even log them. What if `row` is not valid?

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,255 @@
...
+        removeInnerItem(row);
+        endRemoveRows();
+    }
+
+    void PresetKeywordsModel::addItem() {
+        int lastIndex = (int)getPresetsCount();
+        Common::Hold placeHolder;
```
Ribtoks: What is this? Why do you declare local variable which will be destructed after the scope exit? `Common::Hold` does not have a move constructor and should not have one. It will work though if you will allocate it with `new` inside `shared_ptr`

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.h
```diff
@@ -0,0 +1,84 @@
...
+
+#include "../Common/basickeywordsmodel.h"
+#include "../Common/baseentity.h"
+#include "../Common/abstractlistmodel.h"
+#include <QAbstractListModel>
+#include <QVector>
+#include <QSortFilterProxyModel>
+
+namespace Presets {
+    struct Preset {
```
Ribtoks: It is always better to use at least 2 nouns in a name of an entity. Like `KeywordsPreset`

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.h
```diff
@@ -0,0 +1,84 @@
...
+
+    class PresetKeywordsModel:
+        public QAbstractListModel,
+        public Common::BaseEntity
+    {
+    Q_OBJECT
+
+    public:
+        PresetKeywordsModel(QObject *parent=0);
+        size_t getPresetsCount() const {return m_Presets.size(); }
```
Ribtoks: http://doc.qt.io/qt-5/qvector.html#size

`size()` of QVector returns `int`, not `size_t`. It's only `std::vector` has `size_t`.

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.h
```diff
@@ -0,0 +1,84 @@
...
+        Q_INVOKABLE void appendKeyword(int index, const QString &keyword);
+        Q_INVOKABLE QObject *getKeywordsModel(int index);
+        Q_INVOKABLE void saveToConfig();
+        Q_INVOKABLE void loadFromConfigModel();
+
+    public slots:
+        void onPresetsUpdated() { LOG_INFO << "loading Model"; loadFromConfigModel(); }
+
+    private:
+        QVector<Preset> m_Presets;
```
Ribtoks: Would be nice to use at least 2 nouns in the name of this vector. Like "m_PresetsList" or something like that

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.h
```diff
@@ -0,0 +1,84 @@
...
+
+    public:
+        Q_INVOKABLE void removeItem(int row);
+        Q_INVOKABLE void addItem();
+        Q_INVOKABLE void editKeyword(int index, int keywordIndex, const QString &replacement);
+        Q_INVOKABLE QString removeKeywordAt(int index, int keywordIndex);
+        Q_INVOKABLE void appendKeyword(int index, const QString &keyword);
+        Q_INVOKABLE QObject *getKeywordsModel(int index);
+        Q_INVOKABLE void saveToConfig();
+        Q_INVOKABLE void loadFromConfigModel();
```
Ribtoks: Why this method is Q_INVOKABLE? I did not find any references from .qml files (or maybe github did not expand some new ones)

src/xpiks-qt/PresetKeywords/presetkeywordsmodelconfig.cpp
```diff
@@ -0,0 +1,176 @@
...
+            overwrite = overwriteLocal;
+        }
+
+        Models::AbstractConfigUpdaterModel::processRemoteConfig(remoteDocument, overwrite);
+    }
+
+    int PresetKeywordsModelConfig::operator ()(const QJsonObject &val1, const QJsonObject &val2) {
+        Q_UNUSED(val1);
+        Q_UNUSED(val2);
+        return 0;
```
Ribtoks: Can you please leave a small comment "for future generations" explaining what will happen in case of two json objects will be always considered equal

src/xpiks-qt/PresetKeywords/presetkeywordsmodelconfig.h
```diff
@@ -0,0 +1,49 @@
+#ifndef PRESETSKEYWORDSMODELCONFIG_H
+#define PRESETSKEYWORDSMODELCONFIG_H
+
+#include "../Models/abstractconfigupdatermodel.h"
+#include "presetkeywordsmodel.h"
+
+namespace  Presets {
+    struct PresetData {
+        QStringList keys;
```
Ribtoks: Please follow naming conventions in Xpiks

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.h
RostaTasha: The reference exists https://github.com/RostaTasha/xpiks/blob/207/src/xpiks-qt/Dialogs/presetEditDialog.qml#L49

RostaTasha:
Failing test 23:59:55.483 critical T#2852 PlainTextEditTest::doTest - Keywords spell error not detected.
I doubt that this failture could be caused by my changes. And on Lin everything is OK. maybe timing issue?

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -487,6 +493,28 @@ namespace Common {
...
+        }
+
+        LOG_DEBUG << "target word index" << keywordsIndex << " " << keywordsList;
+        QString removedKeyword;
+        bool wasCorrect = false;
+
+        beginRemoveRows(QModelIndex(), keywordsIndex, keywordsIndex);
+        this->takeKeywordAtUnsafe(keywordsIndex, removedKeyword, wasCorrect);
+        endRemoveRows();
+        Q_UNUSED(removedKeyword);
```
Ribtoks: Would be useful to log this removedKeyword with INFO - not just ignore it

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -487,6 +493,28 @@ namespace Common {
         return anythingRemoved;
     }
 
+    bool BasicKeywordsModel::expandPresetUnsafe(int keywordsIndex, const QStringList &keywordsList) {
```
Ribtoks: isn't it a `keyword Index`, not `keywords Index`?

src/xpiks-qt/Models/abstractconfigupdatermodel.cpp
```diff
@@ -58,4 +50,18 @@ namespace Models {
...
     }
+
+    void AbstractConfigUpdaterModel::initRemoteConfig(const QString &configUrl)
+    {
+        Q_ASSERT(m_CommandManager != NULL);
+        Models::SettingsModel *settingsModel = m_CommandManager->getSettingsModel();
+        ProxySettings *proxySettings = settingsModel->retrieveProxySettings();
+        Q_ASSERT(!configUrl.isEmpty());
+        m_RemoteConfig.requestInitConfig(configUrl, proxySettings);
+    }
```
Ribtoks: Please separate different methods implementations with an empty line

src/xpiks-qt/Models/combinedartworksmodel.cpp
```diff
@@ -274,6 +274,32 @@ namespace Models {
         return getHasDescriptionWordSpellError(word);
     }
 
+    void CombinedArtworksModel::replaceFromPreset(int keywordsIndex, int presetIndex) {
+        auto *presetsModel = m_CommandManager->getPresetsModel();
```
Ribtoks: I'm not sure, but it looks like all this functionality should belong to ArtworkProxyBase class since you will absolutely duplicate this code in CombinedArtworksModel and ArtworkProxy classes.

src/xpiks-qt/Models/combinedartworksmodel.cpp
```diff
@@ -274,6 +274,32 @@ namespace Models {
         return getHasDescriptionWordSpellError(word);
     }
 
+    void CombinedArtworksModel::replaceFromPreset(int keywordsIndex, int presetIndex) {
+        auto *presetsModel = m_CommandManager->getPresetsModel();
+        QStringList keywords;
+
+        if (!presetsModel->tryGetPreset(presetIndex, keywords)) {
```
Ribtoks: Please remove the `return` and organize code like in method below `CombinedArtworksModel::appendFromPreset`

src/xpiks-qt/Models/filteredartitemsproxymodel.cpp
```diff
@@ -393,6 +393,45 @@ namespace Models {
...
+            int originalIndex = getOriginalIndex(index);
+            LOG_INFO << "preset index" << presetIndex << "original index" << originalIndex;
+            auto *presetsModel = m_CommandManager->getPresetsModel();
+            QStringList keywords;
+            if (presetsModel->tryGetPreset(presetIndex, keywords)) {
+                ArtItemsModel *artItemsModel = getArtItemsModel();
+                ArtworkMetadata *metadata = artItemsModel->getArtwork(originalIndex);
+
+                if (metadata != NULL) {
+                    auto *keywordsModel = metadata->getBasicModel();
```
Ribtoks: * This seem to be a bug. You definitely should operate with ArtworkMetadata because after replacing from preset it will (or should!) mark itself as modified.
* You should submit it to spellcheck immediately same as you do for CombinedArtworksModel

src/xpiks-qt/Models/filteredartitemsproxymodel.cpp
```diff
@@ -393,6 +393,45 @@ namespace Models {
...
+            auto *presetsModel = m_CommandManager->getPresetsModel();
+            QString presetName;
+            if (presetsModel->tryGetNameFromIndex(presetIndex, presetName)) {
+                QStringList keywords;
+                if (presetsModel->tryGetPreset(presetIndex, keywords)) {
+                    ArtItemsModel *artItemsModel = getArtItemsModel();
+                    ArtworkMetadata *metadata = artItemsModel->getArtwork(originalIndex);
+
+                    if (metadata != NULL) {
+                        auto *keywordsModel = metadata->getBasicModel();
```
Ribtoks: Same as above please: work with metadata instead of keywordsModel and submit to spellcheck afterwards

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,254 @@
...
+        if (presetIndex < 0 || presetIndex >= m_PresetsList.size()) {
+            return false;
+        }
+
+        keywords = m_PresetsList[presetIndex].m_KeywordsModel->getKeywords();
+        return true;
+    }
+
+    void PresetKeywordsModel::removeItem(int row) {
+        if (row < 0 || row >= getPresetsCount()){
```
Ribtoks: Let's also LOG these actions (with `row`) in the very beginning of the method

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,254 @@
...
+            return;
+        }
+        beginRemoveRows(QModelIndex(), row, row);
```
Ribtoks: Please leave at least 1 line of code after block with parenthesis

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,254 @@
...
+        removeInnerItem(row);
+        endRemoveRows();
+    }
+
+    void PresetKeywordsModel::addItem() {
+        int lastIndex = (int)getPresetsCount();
+        beginInsertRows(QModelIndex(), lastIndex, lastIndex);
+        m_PresetsList.emplace_back("Undefined");
+        endInsertRows();
+        LOG_INFO << "#";
```
Ribtoks: such logging should better be in the beginning of the method since it will be useless in case of crash in the bottom

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,254 @@
...
+        emit dataChanged(this->index(lastIndex-1), this->index(lastIndex-1));
+    }
+
+    void PresetKeywordsModel::editKeyword(int index, int keywordIndex, const QString &replacement) {
+        if (0 <= index && index < getPresetsCount()) {
```
Ribtoks: Let's log all actions and parameters of all these public API of PresetKeywordsModel - just like Combined Model and all other models do with their public API

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,254 @@
...
+            }
+        }
+
+        return keyword;
+    }
+
+    void PresetKeywordsModel::appendKeyword(int index, const QString &keyword) {
+        size_t size = getPresetsCount();
+
+        LOG_DEBUG << "Model for item" << index;
```
Ribtoks: Please use LOG_INFO for logging with parameters

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,254 @@
...
+        if (0 <= index && index < size) {
+            auto &keywords = m_PresetsList[index].m_KeywordsModel;
+            keywords->appendKeyword(keyword);
+        }
+    }
+
+    QObject *PresetKeywordsModel::getKeywordsModel(int index) {
+        size_t size = getPresetsCount();
+
+        LOG_DEBUG << "Model for item" << index;
```
Ribtoks: LOG_INFO - same as above

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+        auto &name = m_PresetsList[row].m_PresetName;
+        auto newName = value.toString();
+        switch (role) {
+            case NameRole:
+                if (name == newName) {
+                    return false;
+                } else {
+                    QString sanitized = newName.simplified();
+                    if (sanitized == "") {
```
Ribtoks: This is very minor, but just for the sake of readability it's better to use nice names like `if (sanitized.isEmpty()) {`

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,254 @@
...
+    int FilteredPresetKeywordsModel::getOriginalIndex(int index)
+    {
+        QModelIndex originalIndex = mapToSource(this->index(index, 0));
+        int row = originalIndex.row();
+
+        return row;
+    }
+
+    QString FilteredPresetKeywordsModel::getName(int index)
+    {
```
Ribtoks: Please put the brace on the same line as method + in the method above

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,254 @@
...
+            return true;
+        }
+
+        PresetKeywordsModel *presetsModel = getPresetsModel();
+        QString name;
+        if (!presetsModel->tryGetNameFromIndex(sourceRow, name)) {
+            return false;
+        }
+
+        return name.contains(m_SearchTerm, Qt::CaseInsensitive);
```
Ribtoks: Let's not return function call - it is bad for debugging! There's a better way to write it:
    
    bool contains = name.contains(m_SearchTerm, Qt::CaseInsensitive);
    return contains;

src/xpiks-qt/PresetKeywords/presetkeywordsmodelconfig.h
```diff
@@ -0,0 +1,49 @@
...
+    struct PresetData {
+        QStringList m_Keys;
+        QString m_Name;
+    };
+
+    class PresetKeywordsModelConfig:
+        public Models::AbstractConfigUpdaterModel
+    {
+    Q_OBJECT
+    friend class PresetKeywordsModel;
```
Ribtoks: Is this really needed? `friend class` is one of bad features in C++ (workaround for incapsulation violation) and we should try to omit them as much as possible (currently it is used in 1 place only and I'm looking forward to remove it)

src/xpiks-qt/StackViews/MainGrid.qml
```diff
@@ -871,10 +895,14 @@ ColumnLayout {
...
-                                                            console.log("Context menu for add word " + rightClickedWord)
-                                                            addWordContextMenu.word = rightClickedWord
-                                                            addWordContextMenu.popup()
+                                                        console.log("Context menu for add word " + rightClickedWord)
+                                                        var showAddToDict = filteredArtItemsModel.hasDescriptionWordSpellError(rowWrapper.delegateIndex, rightClickedWord)
+                                                        wordRightClickMenu.showAddToDict = showAddToDict
+                                                        wordRightClickMenu.word = rightClickedWord
+                                                        wordRightClickMenu.showExpandPreset = false
+                                                        if (wordRightClickMenu.showAddToDict ||
+                                                                wordRightClickMenu.showExpandPreset) {
```
Ribtoks: `showExpandPreset`. You set it to `false` on previous line and check if it's true on this line. Is it just for the explicit visual help or it's a typo?

src/xpiks-qt/StackViews/MainGrid.qml
```diff
@@ -60,12 +60,36 @@ ColumnLayout {
...
+            id : presetSubMenu
+            visible: wordRightClickMenu.showExpandPreset
+            title: i18.n + qsTr("Expand as preset")
+            Instantiator {
+                id : presetsInstantiator
+                model: filteredPresetsModel
+                onObjectAdded: presetSubMenu.insertItem( index, object )
+                onObjectRemoved: presetSubMenu.removeItem( object )
+                delegate: MenuItem {
+                    text: i18.n + qsTr("\"%1\"").arg(filteredPresetsModel.getName(index))
```
Ribtoks: Copy-paste error. No need to use `i18.n` here. Preset names are user-generated. They will not be translated in different languages.

src/xpiks-qt/StackViews/CombinedEditView.qml
```diff
@@ -93,12 +93,34 @@ Rectangle {
...
+            id : presetSubMenu
+            visible: wordRightClickMenu.showExpandPreset
+            title: i18.n + qsTr("Expand as preset")
+            Instantiator {
+                id : presetsInstantiator
+                model: filteredPresetsModel
+                onObjectAdded: presetSubMenu.insertItem( index, object )
+                onObjectRemoved: presetSubMenu.removeItem( object )
+                delegate: MenuItem {
+                    text: i18.n + qsTr("\"%1\"").arg(filteredPresetsModel.getName(index))
```
Ribtoks: Copy-paste error. No need to use `i18.n` here. Preset names are user-generated. They will not be translated in different languages.

src/xpiks-qt/StackViews/ArtworkEditView.qml
```diff
@@ -97,12 +97,34 @@ Rectangle {
...
+            id : presetSubMenu
+            visible: wordRightClickMenu.showExpandPreset
+            title: i18.n + qsTr("Expand as preset")
+            Instantiator {
+                id : presetsInstantiator
+                model: filteredPresetsModel
+                onObjectAdded: presetSubMenu.insertItem( index, object )
+                onObjectRemoved: presetSubMenu.removeItem( object )
+                delegate: MenuItem {
+                    text: i18.n + qsTr("\"%1\"").arg(filteredPresetsModel.getName(index))
```
Ribtoks: Same as below - no need to use `i18.n` and `qsTr()` with preset names - they won't be translatable.

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    PresetKeywordsModel *FilteredPresetKeywordsModel::getPresetsModel() const
+    {
+        QAbstractItemModel *sourceItemModel = sourceModel();
+        PresetKeywordsModel *presetsModel = dynamic_cast<PresetKeywordsModel *>(sourceItemModel);
```
Ribtoks: I guess we should also Q_ASSERT that `presetsModel != nullptr` just to control ourselves (developers)

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+    void FilteredPresetKeywordsModel::setSearchTerm(const QString &value) {
+        LOG_INFO << value;
+        bool anyChangesNeeded = value != m_SearchTerm;
+
+        if (anyChangesNeeded) {
+            m_SearchTerm = value;
+            emit searchTermChanged(value);
+        }
+        invalidateFilter();
```
Ribtoks: Wasn't the purpose of `anyChangesNeeded` **not** to `invalidateFilter()` too often? And you moved `invalidateFilter` out of `if` statement so it is executed every time (even when not needed)

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
```diff
@@ -0,0 +1,254 @@
...
+        return roles;
+    }
+
+    void PresetKeywordsModel::removeInnerItem(int row) {
+        Q_ASSERT(row >= 0 && row < getPresetsCount());
+        m_PresetsList.erase(m_PresetsList.begin() + row);
+    }
+
+    int FilteredPresetKeywordsModel::getOriginalIndex(int index)
+    {
```
Ribtoks: Please put brace on prev line

RostaTasha:
In both cases PlainTextEditTest is failing, so, I guess, I may consider the tests to pass.

src/xpiks-qt/Common/basickeywordsmodel.cpp
RostaTasha: Done

src/xpiks-qt/Common/basickeywordsmodel.cpp
RostaTasha: done

src/xpiks-qt/Models/abstractconfigupdatermodel.cpp
RostaTasha: Done

src/xpiks-qt/StackViews/MainGrid.qml
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodelconfig.h
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
RostaTasha: Done

src/xpiks-qt/StackViews/CombinedEditView.qml
RostaTasha: Done

src/xpiks-qt/StackViews/ArtworkEditView.qml
RostaTasha: Done

src/xpiks-qt/StackViews/MainGrid.qml
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
RostaTasha: Done

src/xpiks-qt/PresetKeywords/presetkeywordsmodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/combinedartworksmodel.cpp
RostaTasha: Done

Issue #379, opened 2017-03-02
User dictionary does not accept words with symbols


Ribtoks:
Can't add a word "24/7" or similar

Ribtoks:
Fixed in #384 
Words with dashes should be split anyway.

PR #384, opened 2017-03-02
[WIP] Don't break words with "/" and "-" for user dictionaries

gnidorah:
For https://github.com/Ribtoks/xpiks/issues/379

1. Do only "/" and "-" need to be allowed in a single word?
2. Will my change make regressions?
[ Can't currently run Integration tests:
21:58:25.325 info T#8786 MetadataIO::ReadingOrchestrator::ReadingOrchestrator - Using 1 threads for 0 items to read
...
21:58:25.326 fatal T#8786 unknown - ASSERT: "!itemsToRead.isEmpty()" in file ../../xpiks-qt/MetadataIO/exiv2readingworker.cpp, line 510
]


Ribtoks:
Hi! First of all, thanks for your contribution!

Speaking about your proposed changes:

- changes in the StyledTextEdit are fine
- changes in the foreachWord() are **not fine**

The latter is quite sensitive area and you can't just change it like that. Entry point to the problem is a constructor of `ModifyUserDictItem`. It calls `Helpers::splitText()`, which is just a wrapper over `Helpers::foreachWord()` which is, in turn, wrapper over `Helpers::foreachPart()`.

A **correct** way to fix this issue is to create a different wrapper over `Helpers::foreachPart()` which will exclude slash and dash. This new function will only be used in constructor of `ModifyUserDictItem`. In this way you will not touch other code and fix the issue.

Ribtoks:
@gnidorah as you can see, your changes broke few tests in the `StringHelpersTests` in the core tests

gnidorah:
@Ribtoks Thank you, fixed! Unfortunately change to `ModifyUserDictItem` was not enough :(

- a change in `Common/basicmetadatamodel.cpp` was needed to make right click context menu work for description and title fields (QML model related?)
- a change in `SpellCheck/spellcheckerrorshighlighter.cpp` was needed to make wrong spelled word hightlight working

P.S.: I've not yet looked at, but should metadata related functions be also cared?

Ribtoks:
@gnidorah Thanks a lot, looks much better now!

Change in `SpellCheckErrorsHighlighter` is absolutely valid.

Though I can't see your point about description and title fields correlation to `BasicMetadataModel` changes. Can you elaborate on that? (right-click menu works for description and title field anyway because of changes in `StyledTextEdit`)

Ribtoks:
@gnidorah Hey. Ok, I've spent some time looking into it and why the tests failed.

So the points:

- Actually we **want** to split words by dash (e.g. multi-core should be split)
- `Helpers::splitText()` is used everywhere (in Warnings Checking, in Highlighting, in merging keywords. So when you do a change in `BasicMetadataModel` quite a few things go out of sync since they rely on `Helpers::splitText()`

What to do:

- Roll back to your initial change in the `Helpers::splitText()` **except** adding dash ('-') as a non-split character. So overall you should patch `Helpers::splitText()` and `StyledTextEdit` by remove '/' from separators

Thanks once again and sorry for any confusion!

P.S. The issue was copied from the claim of a user and I didn't investigate the dash thing initially (that we actually need it).
-

Ribtoks:
@gnidorah Unfortunately correct resolution of the issue requires some more changes to be made. I'm talking about functions `Helpers::isLeftWordBound()` and `Helpers::isRightWordBound()` since they rely on the same separation logic `(isSpace() || isPunct())`. I would suggest creating a local function in the `stringhelper.cpp` file - something like `bool isPunctuation(const QChar &)` and use it in the following functions:

- `isRightWordBound()`
- `isLeftWordBound()`
- `foreachWord()`

Ribtoks:
@gnidorah Looks good to me! 

Also I would kindly ask you to follow braces style specified in the CONTRIBUTING.md (opening brace on previous line for `isPunctuation()` function). Thanks!

gnidorah:
@Ribtoks Done, thank you!

Ribtoks:
@gnidorah Thank you very much! Your fix will make it to the v1.4.1 bugfix release.

gnidorah:
@Ribtoks Thank you! Considering all your great help, perhaps it sounds impudently, but could I ask you to send me some money to my PayPal yourbestfriend@openmailbox.org? I'm really intersted in helping to this project on a payed basis: https://www.linux.org.ru/forum/development/12270969

Ribtoks:
@gnidorah I appreciate your help! But in future this will only work if I will not investigate the problem by myself and tell you a complete plan of what and how to do, but rather guide you and answer the questions in a private chat. Cheers

gnidorah:
@Ribtoks Thank you very very much! I completely agree, I should play fair.

PR #392, opened 2017-03-08
New linux build

Artiom-M:


Ribtoks:
@Artiom-M Please remove unnecessary comments in the `xpiks-qt.pro` file (#config += debug) and reordering of the linking libraries in the same file (names are the same - what is the purpose of changing order?)

Artiom-M:
This is exactly the purpose of the change: the order is important. A library should be passed after another library which it uses. libz is used by Quazip for example. Anyway I have other changes to add relative to Suse and to documentation. So I will remove debug config comment as well

Ribtoks:
Ok, thanks

Issue #374, opened 2017-02-24
Update Hunspell usage to safe API


Ribtoks:
`spellcheckworker.cpp` uses a lot of unsafe API from Hunspell library operating on pointers/zero-separated strings though after upgrading to Hunspell 1.6.0 a new API is present (which operates with std::string and std::vector). This task is to replace all calls to the old API with the new API.

PR #395, opened 2017-03-09
Switch to a new Hunspell API

gnidorah:
Should close https://github.com/Ribtoks/xpiks/issues/374


Ribtoks:
@gnidorah look inside `hunspell.pro` - it expects _hunspell_ to be in the `hunspell-1.6.0/` directory

gnidorah:
@Ribtoks It is, the archive contains hunspell-1.6.0 inside
BTW, wrote you via Skype:
>  Seems that package is not compatible: https://travis-ci.org/gnidorah/xpiks/jobs/209442080
So, looking at that situation with Hunspell library still on 1.4 in Debian and Ubuntu suppose we could just build it from sources as temporary measure?
P.S.: Integration test will fail probably due dictionary not on expected location. I will look into it later

Ribtoks:
@gnidorah downloading/building Hunspell is totally fine, however I don't like your change in the .pro files of tests - you will break Windows build.

Issue #377, opened 2017-03-01
Setting for turning off filtering based on path


Ribtoks:
Currently filter field in the `MainGrid.qml` supports filtering by a whole bunch of data (title, description, keywords, reserved words, filepath etc.). This uses `Helpers::hasSearchMatch()` API which filters based on flags and search term.

This task is to make use of a `SearchFlags::Filepath` flag optional by creating a new Setting in the `SettingsWindow.qml` Behavior Tab with string "Search by path" as well as a field in `SettingsModel`  (https://github.com/Ribtoks/xpiks/wiki/How-to-add-a-new-Setting-to-Xpiks) and provide that flag in `FilteredArtItemsProxyModel::filterAcceptsRow()` as parameter for `Helpers::containsPartsSearch()` method.

gnidorah:
Taking this

PR #399, opened 2017-03-13
Setting for turning off filtering based on path

gnidorah:
#377
Thank you!


src/xpiks-qt/Helpers/filterhelpers.cpp
```diff
@@ -194,10 +194,14 @@ namespace Helpers {
         return hasMatch;
     }
 
-    bool containsPartsSearch(const QString &mainSearchTerm, Models::ArtworkMetadata *metadata, bool searchUsingAnd) {
+    bool containsPartsSearch(const QString &mainSearchTerm, Models::ArtworkMetadata *metadata, bool searchUsingAnd, bool searchByFilepath) {
         // default search is not case sensitive
         Common::SearchFlags searchFlags = searchUsingAnd ? Common::SearchFlags::AllTermsEverything :
                                            Common::SearchFlags::AnyTermsEverything;
+        if (!searchByFilepath) {
+          Common::UnsetFlag(searchFlags, Common::SearchFlags::Filepath);
```
Ribtoks: There's a whitespace misalignment here - some spaces are missing.

src/xpiks-qt/Dialogs/SettingsWindow.qml
```diff
@@ -301,6 +301,21 @@ ApplicationWindow {
                             }
                         }
 
+                        StyledCheckbox {
+                            id: searchByFilepathCheckbox
+                            text: i18.n + qsTr("Search by path")
+                            onCheckedChanged: {
+                                settingsModel.searchByFilepath = checked
```
Ribtoks: What if there were some artworks in the main grid already filtered? After changing this setting they should be updated (by calling `invalidateFilter()` of FilteredArtworksModel). I propose to create a signal in SettingsModel smth like `settingsSaved()` or `settingsUpdated()` and slot in FilteredArtworksModel smth like `onSettingsUpdated()` which will `invalidateFilter()` as well as update Filter Flags - pls see below on this.

src/xpiks-qt/Models/filteredartitemsproxymodel.cpp
```diff
@@ -639,8 +639,9 @@ namespace Models {
         if (metadata != NULL) {
             SettingsModel *settingsModel = m_CommandManager->getSettingsModel();
             bool searchUsingAnd = settingsModel->getSearchUsingAnd();
+            bool searchByFilepath = settingsModel->getSearchByFilepath();
 
-            hasMatch = Helpers::containsPartsSearch(m_SearchTerm, metadata, searchUsingAnd);
+            hasMatch = Helpers::containsPartsSearch(m_SearchTerm, metadata, searchUsingAnd, searchByFilepath);
```
Ribtoks: This was fine with 1 boolean, but now it grows and requires a lot of changes (look at your changes in tests). For this to be usable in future (maybe more booleans?), I propose to remove this method `Helpers::containsPartsSearch()` and use only `Helpers::hasSearchMatch()` directly. 

In order to do so, you need:

- add SearchFlags filed to _FilteredArtItemsProxyModel_ (along with m_SearchTerm)
- in the `onSettingsUpdated()` and `setSearchTerm()` you will need to reconstruct it (set/unset different flags)
- in the `FilteredArtItemsProxyModel::filterAcceptsRow()` use directly the flags and `Helpers::hasSearchMatch()` - there will be performance benefits over current setup
- fix tests (in order not to create flags all of the times - create 2 helper methods of same `ArtworkFilterTests` class which will return needed combinations of flags and use the methods directly in the `hasSearchMatch()` call (with which you will need to replace current `Helpers::containsPartsSearch`)

src/xpiks-qt/Models/filteredartitemsproxymodel.cpp
```diff
@@ -54,6 +55,13 @@ namespace Models {
             emit searchTermChanged(value);
         }
 
+        SettingsModel *settingsModel = m_CommandManager->getSettingsModel();
+        bool searchUsingAnd = settingsModel->getSearchUsingAnd();
+        bool searchByFilepath = settingsModel->getSearchByFilepath();
+        // default search is not case sensitive
+        Common::ApplyFlag(m_SearchFlags, searchUsingAnd, Common::SearchFlags::AllTerms);
```
Ribtoks: This is not entirely correct way to do it. The correct one was in the removed `containsParts()`

    Common::SearchFlags searchFlags = searchUsingAnd ? Common::SearchFlags::AllTermsEverything :
                                        Common::SearchFlags::AnyTermsEverything;

src/xpiks-qt/Models/filteredartitemsproxymodel.cpp
```diff
@@ -466,6 +474,15 @@ namespace Models {
         }
     }
 
+    void FilteredArtItemsProxyModel::onSettingsUpdated() {
+        LOG_DEBUG << "#";
+        SettingsModel *settingsModel = m_CommandManager->getSettingsModel();
```
Ribtoks: This is a duplicating code with the `setSearchTerm()`. A better way to do it is to create a method `updateSearchFlags()` or smth like that, put this code in there and then just use it in `onSettingsUpdated()` and `setSearchTerm()`

src/xpiks-qt/Dialogs/SettingsWindow.qml
```diff
@@ -301,6 +301,22 @@ ApplicationWindow {
                             }
                         }
 
+                        StyledCheckbox {
+                            id: searchByFilepathCheckbox
+                            text: i18.n + qsTr("Search by path")
+                            onCheckedChanged: {
+                                settingsModel.searchByFilepath = checked
+                                settingsModel.updateSettings()
```
Ribtoks: This is not a correct place to use this. What if the user just checked/unchecked the checkbox? See my comment in the settingsmodel.cpp file

src/xpiks-qt/Models/settingsmodel.h
```diff
@@ -81,6 +82,7 @@ namespace Models {
     public:
         Q_INVOKABLE void resetAllValues();
         Q_INVOKABLE void saveAllValues();
+        Q_INVOKABLE void updateSettings() { emit settingsUpdated(); }
```
Ribtoks: It doesn't seem to me that this should be a separate method. Rather the signal should be emited the `saveAllValues()`

gnidorah:
@Ribtoks Done. Sorry for making wrong assumptions!

src/xpiks-tests/xpiks-tests-core/artworkfilter_tests.h
```diff
@@ -3,15 +3,25 @@
 
 #include <QObject>
 #include <QtTest/QtTest>
+#include "../../xpiks-qt/Common/flags.h"
 
 class ArtworkFilterTests : public QObject
 {
     Q_OBJECT
+private:
+    Common::SearchFlags flagsAnyFilepath() { return Common::SearchFlags::AnyTermsEverything; }
```
Ribtoks: Can you please rename `flagsAnyFilepath()` to `flagsAnyTermWithFilepath()` and `flagsAnd()` to `flagsAllTermsWithoutFilepath()`

Ribtoks:
@gnidorah All good, but you removed initialization code for `m_SearchFlags` in the Filtered Model's constructor. Have you tried launching Xpiks with your changes and searching by filepath? Please initialize it to `Common::SearchFlags::AnyTermsEverything` in the constructor.

Issue #367, opened 2017-02-12
Copy to quickbuffer from keywords suggestion


Ribtoks:
Right-click menu for the suggested image with copy to quickbuffer action which will copy title/description/keywords.

This requires reading Shutterstock/Fotolia/iStock Microstock API documentation of how to get Title and Description additionally to Keywords. Requires patching of all `*QueryEngine` classes in `Suggestion/` directory.

gnidorah:
Working on it

Ribtoks:
@gnidorah 

Shutterstock API https://developers.shutterstock.com/api/v2
Fotolia API https://us.fotolia.com/Services/API/Rest/Documentation#
iStock/Getty is not public

gnidorah:
@Ribtoks Thank you!

Ribtoks:
@gnidorah My pleasure. I've updated link to Getty's page. Good luck.

PR #404, opened 2017-03-16
Copy to quickbuffer from keywords suggestion

gnidorah:
#367 
Thank you!


src/xpiks-qt/Suggestion/gettyqueryengine.cpp
```diff
@@ -127,8 +127,23 @@ namespace Suggestion {
                     parseKeywords(item["keywords"], keywords);
                 }
 
+                QString title;
+                QString description;
+
+                if (item.contains("title")) {
+                    QJsonValue value = item["title"];
+                    if (value.isString())
```
Ribtoks: Please always use brackets `{ }` even for `if` statements with one line of code. Thank you.

P.S. This is sort of coding style used everywhere so it's best to comply with it.

src/xpiks-qt/Suggestion/gettyqueryengine.cpp
```diff
@@ -127,8 +127,23 @@ namespace Suggestion {
...
+                         title = value.toString();
+                }
+
+                if (item.contains("caption")) {
+                    QJsonValue value = item["caption"];
+                    if (value.isString())
```
Ribtoks: Same here for brackets `{ }`

src/xpiks-qt/QuickBuffer/quickbuffer.cpp
```diff
@@ -205,4 +205,19 @@ namespace QuickBuffer {
 
         emit isEmptyChanged();
     }
+
+    void QuickBuffer::setFromSuggestionArtwork(const std::shared_ptr<Suggestion::SuggestionArtwork> &from) {
+        LOG_DEBUG << "#";
+        Q_ASSERT(from != nullptr);
+
+        auto title = from->getTitle();
```
Ribtoks: If the return type would be a reference, you could write `auto &title` and avoid 1 copying happening here (although Qt Strings are copy-on-write, but this is just for general sense).

Please see the comment for `SuggestionArtwork` class

gnidorah:
Found an issue with old field copying, looking at it

Ribtoks:
@gnidorah No problem, but can you please describe the issue in more details? (or preset steps to reproduce)

gnidorah:
@Ribtoks 
Sure. If I got the APIs right, Fotolia doesn't provide description, and Shutterstock doesn't provide a title.
So when filling from Fotolia or Shutterstock, if there was already a title or description in quickbuffer left from a previous copy to quickbuffer, it will be used on a fill from these.
I will push a possible solution. Probably far from ideal... Let me know.

Ribtoks:
@gnidorah Yes, this situation is tricky overall. The situation is a bit different from`QuickBuffer::setFromBasicModel()` and `QuickBuffer::setFromCurrentEditable()`, where you just want _not to touch everything_. I agree that the logic of your fix can be left as is for now, but the implementation is rather misleading. Too much code doing same thing. You just don't need any `if` statements for `title` and `description` instead of checking if it's empty 2 times in a row.

src/xpiks-qt/QuickBuffer/quickbuffer.cpp
```diff
@@ -205,4 +205,22 @@ namespace QuickBuffer {
...
+        auto description = from->getDescription();
+        auto keywords = from->getKeywordsSet().toList();
+
+        if (title.isNull()) { this->setTitle(QString("")); }
```
Ribtoks: 1. in general `isEmpty()` should be used instead of `isNull()`
2. You don't need `if` statement at all. You can just have `this->setTitle(title);` and that's all.

Issue #376, opened 2017-03-01
Move Qt settings to custom json file


Ribtoks:
Refactor use of all Qt's settings to json file in the appdata directory. 

- Loading/Saving code should be only changed inside SettingsModel class, introducing `camelCase` names for json keys accordingly to the old property names. 
- There should be a different config for release and debug settings (like `debug_settings.json` and `settings.json`
- This touches a lot of code in UI as well (all uses of `appSettings` in the UI code should be refactored to usual usage of `settingsModel` which is already present there) and old stuff should be removed
- `constants.h` should contain new json keys instead of old uppercase keys
- Serialized data (like recent menu entries, microstocks credentials etc.) should be serialized as base64 string, QByteArray via QVariant or smth like that - as raw binary data

gnidorah:
Working on it

PR #408, opened 2017-03-22
Move Qt settings to custom json file

gnidorah:
#376 
Thank you!


src/xpiks-qt/Dialogs/SettingsWindow.qml
```diff
@@ -57,8 +57,8 @@ ApplicationWindow {
 
     function onMasterPasswordSet() {
         console.log('UI:SettingsWindow # Master password changed')
-        appSettings.setValue(appSettings.masterPasswordHashKey, secretsManager.getMasterPasswordHash())
-        appSettings.setValue(appSettings.mustUseMasterPasswordKey, true)
+        settingsModel.setValue(settingsModel.masterPasswordHashKey, secretsManager.getMasterPasswordHash())
```
Ribtoks: If everything lives inside `SettingsModel`, there's some sense in creating public API like `setMasterPasswordHash()` that will do this stuff internally. No need to care about "keys" and "set/get Value()" stuff. This is already done, e.g. in `saveCurrentVersion()` - you have public API which handles the keys internally

src/xpiks-qt/Helpers/localconfig.cpp
```diff
@@ -43,6 +43,10 @@ namespace Helpers {
     void LocalConfig::saveToFile() {
         QFile file(m_FilePath);
 
+        if (m_FilePath.isEmpty()) {
+            return;
```
Ribtoks: I guess you should `Q_ASSERT` that filename is not empty otherwise there's some inconsistency

src/xpiks-qt/Models/settingsmodel.h
```diff
@@ -68,13 +73,35 @@ namespace Models {
         Q_PROPERTY(bool autoCacheImages READ getAutoCacheImages WRITE setAutoCacheImages NOTIFY autoCacheImagesChanged)
         Q_PROPERTY(int artworkEditRightPaneWidth READ getArtworkEditRightPaneWidth WRITE setArtworkEditRightPaneWidth NOTIFY artworkEditRightPaneWidthChanged)
 
+        Q_PROPERTY(QString mustUseMasterPasswordKey READ getMustUseMasterPasswordKey CONSTANT)
```
Ribtoks: Again, no need for ALL these keys since it's better to handle privately without knowing there're some keys

src/xpiks-qt/SpellCheck/spellcheckworker.cpp
```diff
@@ -77,8 +77,9 @@ namespace SpellCheck {
         dicPath = resourcesDir.absoluteFilePath(EN_HUNSPELL_DIC);
 
 #else
-        Helpers::AppSettings appSettings;
-        resourcesPath = appSettings.value(Constants::DICT_PATH, "").toString();
+        if (m_SettingsModel) {
+            resourcesPath = m_SettingsModel->value(Constants::dictPath, "").toString();
+        }
```
Ribtoks: Please add 1 empty line after block

src/xpiks-qt/main.cpp
```diff
@@ -154,12 +153,10 @@ void initQSettings() {
                                             appVersion.left(10));
 }
 
-void ensureUserIdExists(Helpers::AppSettings *settings) {
-    QLatin1String userIdKey = QLatin1String(Constants::USER_AGENT_ID);
-
-    if (!settings->contains(userIdKey)) {
+void ensureUserIdExists(Models::SettingsModel *settings) {
+    if (settings->value(Constants::userAgentId).isNull()) {
```
Ribtoks: It is better to refactor this (to hide keys stuff inside SettingsModel) so this will look like `settings->hasUserAgentId()` instead of need to work with the keys **if this call is used in more than 1 place**

src/xpiks-qt/Models/settingsmodel.cpp
```diff
@@ -32,6 +32,11 @@
 #else
 #  define DEFAULT_EXIFTOOL "exiftool"
 #endif
+#ifdef QT_DEBUG
+#define SETTINGS_FILE "debug_settings.json"
```
Ribtoks: Look, you have introduced both debug-prefixed keys and debug settings file. You need only one of those. If I have debug settings file, I would like the keys to be the same. Or to have 1 file with the different keys. I think former option (2 files with same keys) is a better one.

(but actually you need 3 files - normal, debug and for the integration tests)

src/xpiks-qt/Models/settingsmodel.cpp
```diff
@@ -101,16 +106,44 @@ namespace Models {
...
+            QDir appDataDir(appDataPath);
+            localConfigPath = appDataDir.filePath(SETTINGS_FILE);
+        } else {
+            localConfigPath = SETTINGS_FILE;
+        }
+
+        m_Config.initConfig(localConfigPath);
+
+        QJsonDocument &doc = m_Config.getConfig();
+        if (doc.isObject()) {
```
Ribtoks: What if it's not? The least we can do is to `LOG_WARNING` that something went quite wrong

src/xpiks-qt/Models/settingsmodel.h
```diff
@@ -92,6 +119,189 @@ namespace Models {
         Q_INVOKABLE void saveSelectedDictionaryIndex();
 
     public:
+        Q_INVOKABLE inline void setValue(const QString &key, const QVariant &value) {
```
Ribtoks: As explained in the other comments, this guy and his brothers should be `private` and **not** `Q_INVOKABLE`

src/xpiks-qt/Models/settingsmodel.h
```diff
@@ -68,13 +73,35 @@ namespace Models {
...
+        Q_PROPERTY(QString appVersion READ getAppVersion CONSTANT)
+        QString getAppVersion() const { return QCoreApplication::applicationVersion(); }
+
+        Q_PROPERTY(QString useConfirmationDialogsKey READ getUseConfirmationDialogsKey CONSTANT)
+        QString getUseConfirmationDialogsKey() const { return QLatin1String(Constants::useConfirmationDialogs); }
+
     public:
         explicit SettingsModel(QObject *parent = 0);
-        virtual ~SettingsModel() {}
+        virtual ~SettingsModel() { sync(); }
```
Ribtoks: What if I changed some settings and then the application crashed or the operating system crashed? The result - settings will not be saved. 

What should be done is that settings should be saved as before! In the `saveAllValues()`, `resetToDefault()` and other methods of `SettingsModel` where `AppSettings` class was instantiated and it's destructor invoked (causing settings to be saved)

src/xpiks-qt/main.cpp
gnidorah: @Ribtoks These are the only uses:
./src/xpiks-qt/main.cpp:    if (settings->value(Constants::userAgentId).isNull()) {
./src/xpiks-qt/main.cpp:        settings->setValue(Constants::userAgentId, uuid.toString());
./src/xpiks-qt/main.cpp:    QString userId = settingsModel.value(Constants::userAgentId).toString();
And all seem to use Constants::userAgentId in a different way

src/xpiks-qt/Models/settingsmodel.h
gnidorah: Change added

src/xpiks-qt/SpellCheck/spellcheckworker.cpp
gnidorah: Change added

src/xpiks-qt/Models/settingsmodel.cpp
gnidorah: Change added

src/xpiks-qt/Models/settingsmodel.cpp
gnidorah: Change added

gnidorah:
@Ribtoks 
> If everything lives inside SettingsModel, there's some sense in creating public API like setMasterPasswordHash() that will do this stuff internally. No need to care about "keys" and "set/get Value()" stuff.

> As explained in the other comments, this guy and his brothers should be private and not Q_INVOKABLE

So this means all other places (please see current https://github.com/Ribtoks/xpiks/pull/408/files) should be also switched to this?

Some public API methods for getting/setting values in `QJsonObject` are already dupes of public getters/setters operating on private `SettingsModel` fields - anything to do with this or better not to?

src/xpiks-qt/main.cpp
Ribtoks: @gnidorah First of all, please stop using `QString::isNull()` in favor of `QString::isEmpty`

Secondly, yes, it's better to refactor these, since `value()` and `setValue()` are supposed to become private citizens.

Ribtoks:
@gnidorah Yes, that means other places need to be touched too. Please specify which exactly getter/setters you mean.

gnidorah:
@Ribtoks For example in old code there was
```c++
public:
QString AppSettings::getMustUseMasterPasswordKey() const { return QLatin1String(Constants::USE_MASTER_PASSWORD); }
```
which was used everywhere and
```c++
public:
bool SettingsModel::getMustUseMasterPassword() const { return m_MustUseMasterPassword; }
```
which still not used at all.

In new code we add 
```c++
public:
        Q_INVOKABLE void setUseMasterPassword(bool value) {
            setValue(Constants::useMasterPassword, value);
        }
```
to replace `appSettings.setValue(appSettings.mustUseMasterPasswordKey, true)` while there is
```c++
public:
        void setMustUseMasterPassword(bool mustUseMasterPassword) {
            if (m_MustUseMasterPassword == mustUseMasterPassword)
                return;

            m_MustUseMasterPassword = mustUseMasterPassword;
            emit mustUseMasterPasswordChanged(mustUseMasterPassword);
        }
```
etc

Ribtoks:
@gnidorah `SettingsModel::setMustUseMasterPassword(bool)` should have been used in SettingsWindow in the _Security_ tab.

The guy who has `setValue(Constants::useMasterPassword, value);` should be used instead of `AppSettings`' appropriate method.

And yes, we can/must check/clean stuff, but **very** carefully. I told you this change will touch too many places not to cause bugs.

gnidorah:
@Ribtoks Sorry for delay!

src/xpiks-tests/xpiks-tests-core/filteredmodel_tests.cpp
```diff
@@ -18,6 +18,12 @@
     commandManagerMock.InjectDependency(&filteredItemsModel);\
     commandManagerMock.generateAndAddArtworks(count);
 
+#define GENERATE_AND_INITIALIZE_MODELS(value) \
```
Ribtoks: Why can't you add this to the `DECLARE_MODELS_AND_GENERATE`? Why do you need a separate define?

src/xpiks-qt/Conectivity/updateservice.cpp
```diff
@@ -83,17 +82,16 @@ namespace Conectivity {
...
 
-        int availableValue = appSettings.intValue(appSettings.getAvailableUpdateVersionKey(), 0);
+        int availableValue = m_SettingsModel->getAvailableUpdateVersion();
 
         if ((0 < availableValue) && (availableValue <= XPIKS_VERSION_INT)) {
             LOG_DEBUG << "Flushing available update settings values";
-            appSettings.setValue(appSettings.getPathToUpdateKey(), "");
-            appSettings.setValue(appSettings.getAvailableUpdateVersionKey(), 0);
+            m_SettingsModel->setPathToUpdate("");
+            m_SettingsModel->setAvailableUpdateVersion(0);
```
Ribtoks: Would be nice to forcefully sync settings after this. You can make a public wrapper over `sync()`.

src/xpiks-qt/Dialogs/SettingsWindow.qml
```diff
@@ -57,8 +57,8 @@ ApplicationWindow {
 
     function onMasterPasswordSet() {
         console.log('UI:SettingsWindow # Master password changed')
-        appSettings.setValue(appSettings.masterPasswordHashKey, secretsManager.getMasterPasswordHash())
-        appSettings.setValue(appSettings.mustUseMasterPasswordKey, true)
+        settingsModel.setMasterPasswordHash()
```
Ribtoks: I guess you should refactor out all these to 1 slot of `SettingsModel` and handle it in code behind (c++) instead of in here. I guess it is there for historical reasons - feel free to get rid of it.

src/xpiks-qt/Models/settingsmodel.h
```diff
@@ -91,6 +106,268 @@ namespace Models {
...
+            return value(Constants::uploadHosts).toString();
+        }
+
+        Q_INVOKABLE void saveUploadHosts() {
+            UploadInfoRepository *uploadInfoRepository = m_CommandManager->getUploadInfoRepository();
+
+            setValue(Constants::uploadHosts, uploadInfoRepository->getInfoString());
+        }
+
+        Q_INVOKABLE QString getRecentDirectories() {
```
Ribtoks: I guess all these `set*` wrappers should have at least `LOG_DEBUG << "#"`;

PR #414, opened 2017-03-28
Fix release build

gnidorah:
#376
Thank you!


src/xpiks-qt/Models/settingsmodel.h
```diff
@@ -262,6 +262,7 @@ namespace Models {
         Q_INVOKABLE QString getMasterPasswordHash() { return stringValue(Constants::masterPasswordHash); }
         Q_INVOKABLE QString getDictPath() { return stringValue(Constants::dictPath); }
         Q_INVOKABLE bool getMustUseConfirmationDialogs() { return boolValue(Constants::useConfirmationDialogs, true); }
+        Q_INVOKABLE bool getTelemetryStatus() { return boolValue(Constants::userStatistics, true); }
```
Ribtoks: Why this is `Q_INVOKABLE`? This attribute is needed only for methods which are being called from QML. This one is code behind (c++) only

Ribtoks:
Sorry, fixed myself along with other cleaning.

src/xpiks-qt/Models/settingsmodel.h
gnidorah: @Ribtoks Sorry, just copy-paste. I understand the purpose of `Q_INVOKABLE`

Issue #413, opened 2017-03-28
Refactor system tasks to service and worker


Ribtoks:
Refactor Logs cleaning functionality, settings updating functionality and others to a separate service and lowest-priority thread worker.

gnidorah:
Taking this task

PR #415, opened 2017-04-03
[WIP] Refactor system tasks to service and worker

gnidorah:
#413 
Big sorry for delay!


Ribtoks:
@gnidorah There're certain merge issues. Please fix them and also don't forget to write **correct** relative paths for includes in tests. I mean paths starting with "../../" instead of "../" for all "xpiks-qt/" files.

src/xpiks-qt/Maintenance/addtolibraryjobitem.cpp
```diff
@@ -0,0 +1,71 @@
...
+        m_ArtworksList(artworksList),
+        m_LocalLibrary(localLibrary)
+    {
+
+    }
+
+    void AddToLibraryJobItem::doAddToLibrary() {
+        int length = m_ArtworksList.length();
+
+        LOG_DEBUG << length << "file(s)";
```
Ribtoks: Please use `LOG_INFO` for logging with any parameters.

src/xpiks-qt/Maintenance/checkcredentialsjobitem.cpp
```diff
@@ -19,27 +19,32 @@
...
-#include "../Common/defines.h"
-#include "../Models/proxysettings.h"
-#include "../Helpers/stringhelper.h"
+#include "../Conectivity/ftphelpers.h"
+#include "checkcredentialsjobitem.h"
 
-namespace Conectivity {
-    static size_t throw_away(void *ptr, size_t size, size_t nmemb, void *data) {
+namespace Maintenance {
+    CheckCredentialsJobItem::CheckCredentialsJobItem(Conectivity::UploadContext *context, Models::ArtworkUploader *artworkUploader):
```
Ribtoks: [IMPORTANT] **This definitely shouldn't be touched.** Please revert credentials checking to where it was.

src/xpiks-qt/Maintenance/maintenanceservice.cpp
```diff
@@ -0,0 +1,123 @@
...
+        Q_ASSERT(m_MaintenanceWorker);
+        std::shared_ptr<IMaintenanceItem> jobItem(new LaunchExiftoolJobItem(settingsExiftoolPath, coordinator, commandManager));
+        m_MaintenanceWorker->submitItem(jobItem);
+    }
+
+    void MaintenanceService::addInitializeDictionariesTask(Translation::TranslationManager *translationManager) {
+        LOG_DEBUG << "#";
+        Q_ASSERT(m_MaintenanceWorker);
+        std::shared_ptr<IMaintenanceItem> jobItem(new InitializeDictionariesJobItem(translationManager));
+        m_MaintenanceWorker->submitItem(jobItem);
```
Ribtoks: This item has bigger priority than others real "maintenance" items. Please add it to the queue using `submitFirst()` method.

src/xpiks-qt/Maintenance/maintenanceservice.h
```diff
@@ -0,0 +1,64 @@
...
+        void addLogsCleanupTask();
+        void addUpdatesCleanupTask();
+#ifndef CORE_TESTS
+        void addCheckCredentialsTask(Conectivity::UploadContext *context, Models::ArtworkUploader *artworkUploader);
+#endif
+        void addLaunchExiftoolTask(const QString &settingsExiftoolPath, MetadataIO::MetadataIOCoordinator *coordinator, Commands::CommandManager *commandManager);
+        void addInitializeDictionariesTask(Translation::TranslationManager *translationManager);
+        void addToLibraryTask(const QVector<Models::ArtworkMetadata *> artworksList, Suggestion::LocalLibrary *localLibrary);
+
+    public slots:
```
Ribtoks: Why `stopService()` is a slot? Do you connect to it anywhere?

src/xpiks-qt/Maintenance/maintenanceworker.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+    void MaintenanceWorker::processOneItem(std::shared_ptr<IMaintenanceItem> &item) {
+        auto logsCleanupItem = std::dynamic_pointer_cast<LogsCleanupJobItem>(item);
+        auto updatesCleanupItem = std::dynamic_pointer_cast<UpdatesCleanupJobItem>(item);
+        auto checkCredentialsItem = std::dynamic_pointer_cast<CheckCredentialsJobItem>(item);
+        auto launchExiftoolJobItem = std::dynamic_pointer_cast<LaunchExiftoolJobItem>(item);
+        auto initializeDictionariesJobItem = std::dynamic_pointer_cast<InitializeDictionariesJobItem>(item);
+        auto addToLibraryJobItem = std::dynamic_pointer_cast<AddToLibraryJobItem>(item);
+
+        if (logsCleanupItem) {
```
Ribtoks: Ok, this now looks really too much.

Let's create some object-oriented architecture. Please add something like `virtual processJob() = 0;` method to `IMaintenanceItem` class and in here you can just call `item->processJob()`.

The `processJob()` method will be overriden per implementation of the `IMaintenanceItem` and will do inner appropriate operations. Then all this and all code below in this class should be removed.

src/xpiks-qt/Maintenance/launchexiftooljobitem.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+    void LaunchExiftoolJobItem::afterExiftoolDiscovery() {
+        if (!m_MetadataIOCoordinator->getExiftoolNotFound() && !m_RecommendedExiftoolPath.isEmpty()) {
+            LOG_DEBUG << "Recommended exiftool path is" << m_RecommendedExiftoolPath;
+
+            Models::SettingsModel *settingsModel = m_CommandManager->getSettingsModel();
+            QString existingExiftoolPath = settingsModel->getExifToolPath();
+
+            if (existingExiftoolPath != m_RecommendedExiftoolPath) {
+                LOG_INFO << "Setting exiftool path to recommended";
```
Ribtoks: Please use `LOG_DEBUG` for logging without parameters (that string is hardcoded).

src/xpiks-qt/Commands/commandmanager.cpp
```diff
@@ -827,8 +834,12 @@ void Commands::CommandManager::afterConstructionCallback() {
 #endif
 
 #if !defined(CORE_TESTS) && !defined(INTEGRATION_TESTS)
-    Helpers::performCleanLogsAsync();
-    Helpers::cleanupUpdateArtifactsAsync();
+#ifdef WITH_LOGS
+    m_MaintenanceService->addLogsCleanupTask();
+#endif
+#ifdef Q_OS_WIN
```
Ribtoks: Let's add the ifdef inside `MaintenanceService` method so the code here will look way more nicer. The same for other maintenance tasks please

src/xpiks-qt/Commands/commandmanager.cpp
```diff
@@ -401,6 +405,7 @@ void Commands::CommandManager::ensureDependenciesInjected() {
     Q_ASSERT(m_TranslationManager != NULL);
     Q_ASSERT(m_ArtworkProxyModel != NULL);
     Q_ASSERT(m_QuickBuffer != NULL);
+    Q_ASSERT(m_MaintenanceService != NULL);
```
Ribtoks: Please initialize `m_MaintenanceService` to `NULL` in the constructor of `CommandManager` otherwise this check does not make any sense.

src/xpiks-qt/Maintenance/addtolibraryjobitem.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+        QMutexLocker locker(&m_LocalLibrary->getMutex());
```
Ribtoks: This is **very very bad**. Nobody should know LocalLibrary has any sort of Mutex and moreover it **cannot** be exposed outside. This is a recipe for a deadlock.

What you should do is to leave everything as it was - this code should be owned by LocalLibrary and in this `AddToLibraryJobItem` you will just call a method of `LocalLibrary` - absolutely the same way as it was before with `QtConcurrent::Run`

src/xpiks-qt/Maintenance/initializedictionariesjobitem.cpp
```diff
@@ -0,0 +1,89 @@
...
+                ifoFound = ifoFound || suffix.contains("ifo");
+            }
+        }
+
+        bool success = ifoFound && dictFound && indexFound;
+        return success;
+    }
+
+    void InitializeDictionariesJobItem::doInitializeDictionaries() {
+        QString appDataPath = XPIKS_USERDATA_PATH;
```
Ribtoks: All this code is calling back and forth `TranslationManager`'s methods but instead you should have left that in the ownership of the TranslationManager! In here you would just call 1 method of `TranslationManager`.

gnidorah:
@Ribtoks Thank you much for the review!

src/xpiks-qt/Commands/commandmanager.cpp
```diff
@@ -827,8 +834,8 @@ void Commands::CommandManager::afterConstructionCallback() {
 #endif
 
 #if !defined(CORE_TESTS) && !defined(INTEGRATION_TESTS)
-    Helpers::performCleanLogsAsync();
-    Helpers::cleanupUpdateArtifactsAsync();
+    m_MaintenanceService->addLogsCleanupTask();
```
Ribtoks: @gnidorah Maybe it will be nicer to name the method just `cleanupLogs()` and inside the MaintenanceService will add Task?

src/xpiks-qt/Maintenance/addtolibraryjobitem.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include "../Suggestion/locallibrary.h"
+#include "../Models/imageartwork.h"
+#include "addtolibraryjobitem.h"
+
+namespace Maintenance {
+    AddToLibraryJobItem::AddToLibraryJobItem(const QVector<Models::ArtworkMetadata *> artworksList, Suggestion::LocalLibrary *localLibrary):
+        m_ArtworksList(artworksList),
+        m_LocalLibrary(localLibrary)
+    {
+
```
Ribtoks: You should `Q_ASSERT` that `localLibrary` is not null and artworks list is not empty.

src/xpiks-qt/Maintenance/addtolibraryjobitem.h
```diff
@@ -0,0 +1,46 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef ADDTOLIBRARYJOBITEM_H
+#define ADDTOLIBRARYJOBITEM_H
+
+#include <QVector>
+#include "../Models/artworkmetadata.h"
+#include "../Suggestion/locallibrary.h"
```
Ribtoks: There's no need to `include` the whole file `artworkmetadata.h` and `locallibrary.h`. They are used here only like pointers so the compiler does not need to know size of this class, because size of a pointer is always equal to the architecture word (32 or 64 bytes).

Therefore please remove the `includes` - this will speed up the complication process and add forward declarations of these classes like this:

    namespace Models {
        class ArtworkMetadata;
    }

    namespace Suggestion {
        class LocalLibrary;
    }

Please do the same for other files (maintenance tasks) where you did the same.

src/xpiks-qt/Maintenance/initializedictionariesjobitem.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include <QString>
+#include <QDirIterator>
+#include "../Helpers/constants.h"
+#include "initializedictionariesjobitem.h"
+
+namespace Maintenance {
+    InitializeDictionariesJobItem::InitializeDictionariesJobItem(Translation::TranslationManager *translationManager):
+        m_TranslationManager(translationManager)
+    {
+
```
Ribtoks: Also would be nice to have an assert that `translationManager` is not null

src/xpiks-qt/Maintenance/launchexiftooljobitem.h
```diff
@@ -0,0 +1,46 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+#ifndef LAUNCHEXIFTOOLJOBITEM_H
+#define LAUNCHEXIFTOOLJOBITEM_H
+
+#include <QString>
+#include "../MetadataIO/metadataiocoordinator.h"
```
Ribtoks: Same about `include` stuff (read comments above)

src/xpiks-qt/Maintenance/logscleanupjobitem.cpp
```diff
@@ -1,37 +1,50 @@
...
+    }
+
+    void LogsCleanupJobItem::processJob() {
+        LOG_DEBUG << "#";
+        doCleanLogs();
     }
 
-    void cleanLogsLogic(const QString &logsDir) {
+    void LogsCleanupJobItem::doCleanLogs() {
         LOG_DEBUG << "#";
```
Ribtoks: No need to have 2 `LOG_DEBUG` statement in a row - you're logging in a wrapper method above

src/xpiks-qt/Maintenance/maintenanceservice.h
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MAINTENANCESERVICE_H
+#define MAINTENANCESERVICE_H
+
+#include <QObject>
+#include "../MetadataIO/metadataiocoordinator.h"
```
Ribtoks: Same as above regarding includes. Only add them to cpp file and add forward declarations here

src/xpiks-qt/Maintenance/maintenanceworker.cpp
```diff
@@ -0,0 +1,36 @@
...
+namespace Maintenance {
+    bool MaintenanceWorker::initWorker() {
+        LOG_DEBUG << "#";
+        return true;
+    }
+
+    void MaintenanceWorker::processOneItem(std::shared_ptr<IMaintenanceItem> &item) {
+        item->processJob();
+
+        QThread::sleep(5);
```
Ribtoks: Let's refactor `5` to be a constant (smth like `MAINTENANCE_SLEEP`) and set it to `1` instead of `5`. We need to measure real impact on performance of other important background threads (like ImageCaching) and only then set a healthy delay.

src/xpiks-qt/Maintenance/updatescleanupjobitem.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include "updatescleanupjobitem.h"
+
+namespace Maintenance {
+    UpdatesCleanupJobItem::UpdatesCleanupJobItem()
+    {
+
+    }
+
+    void UpdatesCleanupJobItem::processJob() {
+        LOG_DEBUG << "#";
```
Ribtoks: We can add an assert here that this is for Windows only. Something like `#ifndef Q_OS_WIN  Q_ASSERT(false); #endif`

src/xpiks-qt/Suggestion/locallibrary.cpp
```diff
@@ -45,21 +46,20 @@ namespace Suggestion {
...
     LocalLibrary::~LocalLibrary() {
-        delete m_FutureWatcher;
     }
 
     void LocalLibrary::addToLibrary(const QVector<Models::ArtworkMetadata *> artworksList) {
         // adding to library will be complicated in future
         // so always do it in the background
 #ifndef INTEGRATION_TESTS
-        m_FutureWatcher->setFuture(QtConcurrent::run(this, &LocalLibrary::doAddToLibrary, artworksList));
+        Q_ASSERT(m_CommandManager != NULL);
```
Ribtoks: I'm not sure if you need this assert. This should be set up in `CommandManager::InjectDependency`.

src/xpiks-qt/main.cpp
```diff
@@ -298,6 +299,8 @@ int main(int argc, char *argv[]) {
         qunsetenv(highDpiEnvironmentVariable);
     }
 
+    Commands::CommandManager commandManager;
+    localLibrary.setCommandManager(&commandManager);
```
Ribtoks: [important] This is a bad thing to do here. At the time you do `loadLibraryAsync()` (1 line below) you rely on many things set in CommandManager, but in reality they are set a dozen lines below (and due to context switch a background thread might start executing right away and you will run into Access Violation errors because of null pointers). 

So what you should do, is not to move CommandManager here, but to move `localLibrary.loadLibraryAsync();` after line `ensureDependenciesInjected()`. Best of all - move it to `afterConstructionCallback()` - you will be sure everything is properly initialized by that time.

PR #419, opened 2017-04-14
Move Qt settings to custom json file

gnidorah:
#376 
Thank you!


src/xpiks-qt/Models/recentitemsmodel.cpp
```diff
@@ -62,6 +62,12 @@ namespace Models {
         m_RecentItems = deserialized;
     }
 
+    void RecentItemsModel::deserializeFromSettingsAndReset(const QString &serialized) {
```
Ribtoks: ThisNameIsABitTooLongToReadAndUnderstand

src/xpiks-qt/Models/settingsmodel.cpp
```diff
@@ -38,6 +39,8 @@
 #define SETTINGS_FILE "settings.json"
 #endif
 
+#define SETTINGS_VERSION_QSETTINGS_TO_JSON 1
```
Ribtoks: I guess something like `CURRENT_SETTINGS_VERSION` might be better

src/xpiks-qt/Models/settingsmodel.cpp
```diff
@@ -133,6 +136,109 @@ namespace Models {
...
+    void SettingsModel::doMoveSettings() {
+        int settingsVersion = getSettingsVersion();
+
+        if (settingsVersion < SETTINGS_VERSION_QSETTINGS_TO_JSON) {
+            moveSettingsFromQSettingsToJson();
+        }
+    }
+
+    void SettingsModel::moveSetting(const char *oldCharKey, const char *newKey, int type) {
+        QString oldKey = QLatin1String(oldCharKey);
```
Ribtoks: I guess there's no need to do this double conversion. `QLatin1String` constructor from `const char*` and `QString` constructor from `QLatin1String`. Just do it directly..

src/xpiks-qt/Models/settingsmodel.cpp
```diff
@@ -133,6 +136,109 @@ namespace Models {
...
+
+        if (!m_SettingsQSettings->contains(oldKey)) {
```
Ribtoks: Let's `LOG_WARNING` here. Only we own this code so if there will be smth missing - it's a developers bug, not improper user interaction

src/xpiks-qt/Models/settingsmodel.cpp
```diff
@@ -133,6 +136,109 @@ namespace Models {
...
+        auto newValue = QJsonValue::fromVariant(oldValue);
+
+        setValue(newKey, newValue);
+    }
+
+    void SettingsModel::moveProxyHostSetting() {
+        QString oldKey = QLatin1String(Constants::PROXY_HOST);
+
+        if (!m_SettingsQSettings->contains(oldKey)) {
+            return;
```
Ribtoks: Again, `LOG_WARNING` here or anything else to notify world outside

src/xpiks-qt/Models/settingsmodel.cpp
```diff
@@ -133,6 +136,109 @@ namespace Models {
...
+        moveSetting(CACHE_IMAGES_AUTOMATICALLY, cacheImagesAutomatically, QMetaType::Bool);
+        moveSetting(SCROLL_SPEED_SENSIVITY, scrollSpeedSensivity, QMetaType::Double);
+        moveSetting(AUTO_DOWNLOAD_UPDATES, autoDownloadUpdates, QMetaType::Bool);
+        moveSetting(PATH_TO_UPDATE, pathToUpdate, QMetaType::QString);
+        moveSetting(AVAILABLE_UPDATE_VERSION, availableUpdateVersion, QMetaType::Int);
+        moveSetting(ARTWORK_EDIT_RIGHT_PANE_WIDTH, artworkEditRightPaneWidth, QMetaType::Int);
+        moveSetting(TRANSLATOR_SELECTED_DICT_INDEX, translatorSelectedDictIndex, QMetaType::Int);
+
+        // apply imported settings
+        readAllValues();
```
Ribtoks: [important] What is the point of doing `read()` here? Reading will look inside .json file. You have to save moved settings first, because in `readAllValues()` you will overwrite all of them, no?

src/xpiks-qt/Models/settingsmodel.cpp
```diff
@@ -133,6 +136,109 @@ namespace Models {
...
+            return;
+        }
+
+        QVariant qvalue = m_SettingsQSettings->value(oldKey);
+        ProxySettings proxySettings = qvalue.value<ProxySettings>();
+        setValue(Constants::proxyHost, serializeProxyForSettings(proxySettings));
+    }
+
+    void SettingsModel::wipeOldSettings() {
+        m_SettingsQSettings->clear();
```
Ribtoks: We should assert that `m_SettingsQSettings` is not `NULL`

src/xpiks-qt/Models/settingsmodel.h
```diff
@@ -591,6 +603,7 @@ namespace Models {
     private:
         Helpers::LocalConfig m_Config;
         QJsonObject m_SettingsJson;
+        QSettings *m_SettingsQSettings;
```
Ribtoks: If you want to use it as a member, you should initialize it in constructor to `NULL`.

**But** this is a very short-living variable (lifespan of 1 method). You should just better have it local in `moveSettingsFromQSettingsToJson()` and pass as parameter to `moveSettings()`

src/xpiks-qt/Models/settingsmodel.cpp
```diff
@@ -133,6 +136,109 @@ namespace Models {
...
+        oldValue.convert(type);
+        auto newValue = QJsonValue::fromVariant(oldValue);
+
+        setValue(newKey, newValue);
+    }
+
+    void SettingsModel::moveProxyHostSetting() {
+        QString oldKey = QLatin1String(Constants::PROXY_HOST);
+
+        if (!m_SettingsQSettings->contains(oldKey)) {
```
Ribtoks: We should `Q_ASSERT` that `m_SettingsQSettings` is not `NULL`

src/xpiks-qt/main.qml
```diff
@@ -390,6 +395,11 @@ ApplicationWindow {
                 title: i18.n + qsTr("&Recent files")
                 enabled: applicationWindow.openedDialogsCount == 0
 
+                Connections {
+                    target: settingsModel
+                    onSettingsMoved: settingsModel.updateRecentFiles();
```
Ribtoks: You should not do this update and connection. This is a Qt Model - just do beginReset() and endReset() when updating it and that's all. This model *should not know* about moving settings in the first place (which is a responsibility of SettingsModel).

Issue #368, opened 2017-02-12
Ability to remove duplicates of word forms


Ribtoks:
Detect plural forms and similar and remove the duplicate.

End goal is to create 2 UI representations:

1. When you type keywords duplicates are highlighted in blue (like red for incorrect spelling)
2. Dialog to cleanup duplicates (like after clicking "Fix spelling")

Iteration 1:

- [ ] SpellCheckingWorker - add API to get stem using Hunspell API `spell()` - it has out parameter which is not used right now
- [ ] BasicKeywordsModel - add map {stem}->{number of keywords with this stem} and update this mapping when adding/removing keywords (stems will be kept in the struct - read below)
- [ ] BasicKeywordsModel - extend SpellCheckResults to be vector of structs instead of bool's. Struct should contain stem and boolean
- [ ] BasicKeywordsModel / SpellCheckingWorker - extend functionality of the worker to produce stems as well as spellcheck statuses and pass through them to BasicKeywordsModel like spellchecking statuses are passed right now
- [ ] ArtworkMetadata - add method to tell if for this keywords there're other keywords with this stem (like if map from step one for the stem of this keyword contains value > 1)
- [ ] ArtItemsModel - add another role like _hassynonyms_ which will return the result of the previous method from ArtworkMetadata
- [ ] KeywordWrapper.qml - add another property _hasDuplicate_ and underscore in blue (artworkActiveColor) instead of red

Iteration 2:

- [ ] Create model like SuggestionModel 
- [ ] Create UI to show this model. In the UI the task is to select duplicates which will be removed

Ribtoks:
https://github.com/snowballstem/snowball/ - as a next step after this is done

Ribtoks:
b7e9dc39c3ca

PR #427, opened 2017-05-15
closing #368

RostaTasha:


src/xpiks-qt/Models/artworkmetadata.cpp
```diff
@@ -107,6 +107,11 @@ namespace Models {
         return isInDir;
     }
 
+    bool ArtworkMetadata::hasSynonyms(int keywordIndex) const
+    {
+        return m_MetadataModel.getSynonymsCount(keywordIndex) > 1;
```
Ribtoks: Again - same performance stuff like in the method above. Maybe would be better to have `anySynonymsFound()` like for the KeywordModel which will break as soon as synonym found on the contrary to calculating all of them.

src/xpiks-qt/SpellCheck/spellcheckworker.cpp
```diff
@@ -266,6 +267,29 @@ namespace SpellCheck {
...
     }
 
+    QStringList SpellCheckWorker::getWordStem(const QString &word) {
+        QStringList result;
+
+        if (word.size() == 0) {
+            return result;
+        }
+        std::string encodedWord = m_Codec->fromUnicode(word).toStdString();
+        const std::vector<std::string> & stems = m_Hunspell->stem(encodedWord);
```
Ribtoks: There should be a way to get 1 result as a "stem" (because there's only 1 stem of a word) instead of all results. This looks like unnecessary computations in the other parts of the application.

src/xpiks-qt/CollapserTabs/QuickBufferTab.qml
```diff
@@ -354,6 +354,7 @@ ColumnLayout {
                 isHighlighted: false
                 keywordText: keyword
                 hasSpellCheckError: !iscorrect
+                hasDuplicate: hasdublicate
```
RostaTasha: done

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -932,12 +985,51 @@ namespace Common {
...
+        QModelIndex start = this->index(0);
+        QModelIndex end = this->index(m_KeywordsList.length() - 1);
+        emit dataChanged(start, end, QVector<int>() << HasDuplicateRole);
     }
 
     QHash<int, QByteArray> BasicKeywordsModel::roleNames() const {
         QHash<int, QByteArray> roles;
         roles[KeywordRole] = "keyword";
         roles[IsCorrectRole] = "iscorrect";
+        roles[HasDuplicateRole] = "hasdublicate";
```
RostaTasha: Done

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -84,7 +84,9 @@ namespace Common {
             case KeywordRole:
                 return m_KeywordsList.at(index.row());
             case IsCorrectRole:
-                return m_SpellCheckResults.at(index.row());
+                return m_SpellCheckResults.at(index.row()).m_SpellCheck;
```
RostaTasha: Renamed

src/xpiks-qt/SpellCheck/spellcheckworker.cpp
RostaTasha: Done

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -932,12 +985,51 @@ namespace Common {
...
+        }
+    }
+
+    void BasicKeywordsModel::removeStemsData(int index) {
+        QWriteLocker writeLocker(&m_StemsLock);
+
+        if (0 <= index && index < m_SpellCheckResults.length()) {
+            const auto &stems = m_SpellCheckResults[index].m_Stems;
+            LOG_INFO << stems;
+            for (const auto &stem : stems) {
```
RostaTasha: Done

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -667,7 +715,7 @@ namespace Common {
 
         // TODO: use smth like memset
         for (int i = 0; i < size; ++i) {
-            m_SpellCheckResults[i] = true;
+            m_SpellCheckResults[i] = Common::WordAnalysisResult();
```
RostaTasha: done

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -140,11 +167,15 @@ namespace Common {
...
 
+        emitDublicateStateChanged();
+
         return result;
     }
 
     bool BasicKeywordsModel::removeLastKeyword(QString &removedKeyword) {
         bool wasCorrect = false, result = false;
+        int index = m_KeywordsList.length() - 1;
+        removeStemsData(index);
```
RostaTasha: Done

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -125,6 +150,8 @@ namespace Common {
     bool BasicKeywordsModel::removeKeywordAt(int index, QString &removedKeyword) {
         bool wasCorrect = false, result = false;
 
+        removeStemsData(index);
```
RostaTasha: Done

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -84,7 +84,9 @@ namespace Common {
             case KeywordRole:
                 return m_KeywordsList.at(index.row());
             case IsCorrectRole:
-                return m_SpellCheckResults.at(index.row());
+                return m_SpellCheckResults.at(index.row()).m_SpellCheck;
+            case HasDuplicateRole:
+                return getSynonymsCount(index.row()) > 1;
```
RostaTasha: Changed to usage of one stem for one word

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -280,6 +316,11 @@ namespace Common {
         Q_UNUSED(writeLocker);
 
         bool result = removeKeywordsUnsafe(keywords, caseSensitive);
+        if (result) {
+            emitDublicateStateChanged();
```
RostaTasha: It's a typo

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -878,25 +940,29 @@ namespace Common {
...
             if (index >= keywordsLength) {
                 break;
             }
         }
+
+        reevaluateStemsData();
     }
 
-    bool BasicKeywordsModel::isReplacedADuplicateUnsafe(int index, const QString &existingPrev,
+    bool BasicKeywordsModel::isReplacedDuplicateUnsafe(int index, const QString &existingPrev,
```
Ribtoks: Looks like you have touched a lot of wrong places when you did some global Find&Replace in your editor. Can you revert those places which you did not touch intentionally? Like all those logs below etc.

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -932,12 +998,59 @@ namespace Common {
                 emit dataChanged(i, i, QVector<int>() << IsCorrectRole);
             }
         }
+
+        emitDuplicateStateChanged();
+    }
+
+    void BasicKeywordsModel::reevaluateStemsData()
```
Ribtoks: Also sounds like a candidate for a Unsafe/normal methods refactoring.

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -932,12 +998,59 @@ namespace Common {
...
+                m_StemsResults[stem].append(word);
+            }
+            i++;
+        }
+    }
+
+    void BasicKeywordsModel::removeStemsData(int index) {
+        QWriteLocker writeLocker(&m_StemsLock);
+
+        if (0 <= index && index < m_SpellCheckResults.length()) {
```
Ribtoks: And this one as well. First method should acquire a lock, check input data and call Unsafe method.

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -878,25 +940,29 @@ namespace Common {
                 if (m_KeywordsList[index].contains(item->m_Word)) {
                     // if keyword contains several words, there would be
                     // several queryitems and there's error if any has error
-                    m_SpellCheckResults[index] = m_SpellCheckResults[index] && item->m_IsCorrect;
+                    m_SpellCheckResults[index].m_IsCorrect = m_SpellCheckResults[index].m_IsCorrect && item->m_IsCorrect;
+                    // consolidate information about stems
+                    m_SpellCheckResults[index].m_Stems.append(item->m_Stem);
```
Ribtoks: I guess it's not very good to always append Stems. Will be back to you with more details later.

src/xpiks-qt/Models/artitemsmodel.cpp
```diff
@@ -687,6 +687,16 @@ namespace Models {
         }
     }
 
+    bool ArtItemsModel::hasSynonyms(int metadataIndex, int keywordIndex) const
+    {
+        if (0 <= metadataIndex && metadataIndex < getArtworksCount()) {
+            ArtworkMetadata *metadata = m_ArtworkList.at(metadataIndex);
+            return metadata->hasSynonyms(keywordIndex);
```
Ribtoks: This one should be not-const guy who will indeed acquire a lock.

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -114,6 +116,44 @@ namespace Common {
         return m_KeywordsList.join(", ");
     }
 
+    int BasicKeywordsModel::getSynonymsCount(int keywordIndex) const {
+        if ((0 <= keywordIndex) && (keywordIndex < m_SpellCheckResults.length())) {
+            const QStringList &stems = m_SpellCheckResults[keywordIndex].m_Stems;
+            const auto target_words = m_KeywordsList[keywordIndex].split(QChar::Space, QString::SkipEmptyParts);
```
RostaTasha: Done

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -114,6 +116,44 @@ namespace Common {
...
+                if (stem.isEmpty())
+                    continue;
+                const auto & target_word = target_words[i];
+                auto it = m_StemsResults.find(stem);
+                if (it != m_StemsResults.end()) {
+                    auto & words = it.value();
+                    for (const auto & word : words) {
+                        if (Helpers::levensteinDistance(target_word, word) <= SYNONYMS_DISTANCE) {
+                            result++;
+                            if (result > 1) {
```
RostaTasha: Done

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -114,6 +116,44 @@ namespace Common {
...
+                        }
+                    }
+                }
+            }
+
+            LOG_INFO << keywordIndex << stems << result;
+            return result;
+        }
+
+        return 0;
```
RostaTasha: Done

src/xpiks-qt/Common/basickeywordsmodel.cpp
```diff
@@ -291,7 +344,7 @@ namespace Common {
             int keywordsCount = m_KeywordsList.length();
 
             m_KeywordsSet.insert(sanitizedKeyword.toLower());
-            m_SpellCheckResults.append(true);
+            m_SpellCheckResults.append(Common::WordAnalysisResultConsolidated());
```
RostaTasha: changed to WordAnalysisResult

src/xpiks-qt/Common/basickeywordsmodel.cpp
RostaTasha: Done

src/xpiks-qt/Common/basickeywordsmodel.cpp
RostaTasha: Done

src/xpiks-qt/Common/basickeywordsmodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/artitemsmodel.cpp
RostaTasha: I can't say, that I got your point here. I  get synonyms data in qml directly from basickeywords model..

Ribtoks: @RostaTasha Maybe I messed this up with some method in ArtworkMetadata, sorry.

src/xpiks-qt/Common/basickeywordsmodel.h
```diff
@@ -168,6 +180,14 @@ namespace Common {
         bool isReplacedADuplicateUnsafe(int index, const QString &existingPrev,
                                         const QString &replacement) const;
         void emitSpellCheckChanged(int index=-1);
+#ifdef CORE_TESTS
```
Ribtoks: What if it's not `CORE_TESTS`? Then all method below will be accounted as signals..

src/xpiks-qt/Common/basicmetadatamodel.cpp
```diff
@@ -358,23 +358,23 @@ namespace Common {
         emit spellCheckResultsReady();
     }
 
-    void BasicMetadataModel::updateDescriptionSpellErrors(const QHash<QString, bool> &results) {
+    void BasicMetadataModel::updateDescriptionSpellErrors(const QHash<QString, Common::WordAnalysisResult> &results) {
         QSet<QString> descriptionErrors;
         QStringList descriptionWords = getDescriptionWords();
         foreach(const QString &word, descriptionWords) {
-            if (results.value(word, true) == false) {
+            if (results.value(word, Common::WordAnalysisResult()).m_IsCorrect == false) {
```
Ribtoks: 1. So for each iteration of the loop you're instantiating an instance of a class.. I guess it will be nicer to have something like `emptyWordAnalysisResult` and have it as a default.
2. Please do not have function calls in the `if` statements. Ever. Let's better have the call separate and then check the return value. I've stated a lot of times that this incomplicates the debugging.

src/xpiks-qt/Common/basicmetadatamodel.cpp
```diff
@@ -358,23 +358,23 @@ namespace Common {
...
         m_SpellCheckInfo->setDescriptionErrors(descriptionErrors);
     }
 
-    void BasicMetadataModel::updateTitleSpellErrors(const QHash<QString, bool> &results) {
+    void BasicMetadataModel::updateTitleSpellErrors(const QHash<QString, Common::WordAnalysisResult> &results) {
         QSet<QString> titleErrors;
         QStringList titleWords = getTitleWords();
         foreach(const QString &word, titleWords) {
-            if (results.value(word, true) == false) {
+            if (results.value(word, Common::WordAnalysisResult()).m_IsCorrect == false) {
```
Ribtoks: Same instantiation and `if`-call problem as in a comment above.

src/xpiks-qt/Common/defines.h
```diff
@@ -89,5 +89,16 @@
 #define SPECIAL_ID_ARTWORK_PROXY_MODEL 2
 #define SPECIAL_ID_COMBINED_MODEL 3
 
+#if defined(CORE_TESTS)
+#define SYNONYMS_DISTANCE 3
+#else
+#if defined (INTEGRATION_TESTS)
```
Ribtoks: `INTEGRATION_TESTS` should reflect real-world app (just without UI) or contain stress-test approach value. So it might be useful to have same value there or value which leads to _more_ computations. Please make sure 100 times that it is correct.

src/xpiks-qt/Common/wordanalysisresult.h
```diff
@@ -0,0 +1,43 @@
...
+        WordAnalysisResult():
+            m_Stem(), m_IsCorrect(true)
+        {}
+
+        WordAnalysisResult(const QString & stem, bool spellCheck):
+            m_Stem(stem), m_IsCorrect(spellCheck)
+        {}
+
+        void reset()
+        {
```
Ribtoks: Braces on the same line, please..

src/xpiks-qt/Common/wordanalysisresult.h
```diff
@@ -0,0 +1,43 @@
...
+namespace Common {
+    struct WordAnalysisResult {
+        QString m_Stem;
+        bool m_IsCorrect;
+
+        WordAnalysisResult():
+            m_Stem(), m_IsCorrect(true)
+        {}
+
+        WordAnalysisResult(const QString & stem, bool spellCheck):
```
Ribtoks: `*` and `&` next to the variable, please. This is all over this pull request..

src/xpiks-qt/Dialogs/RemoveDuplicatesDialog.qml
```diff
@@ -0,0 +1,227 @@
+/*
+ * This file is a part of Xpiks - cross platform application for
+ * keywording and uploading images for microstocks
+ * Copyright (C) 2014-2017 Taras Kushnir <kushnirTV@gmail.com>
+ *
+ * Xpiks is distributed under the GNU General Public License, version 3.0
```
Ribtoks: I've updated the license to LGPL. Please do a full Search-Replace for a 
"GNU General Public License" to "GNU Lesser General Public License" all over your changes.

src/xpiks-qt/Dialogs/RemoveDuplicatesDialog.qml
```diff
@@ -0,0 +1,227 @@
...
+            Component.onCompleted: anchors.centerIn = undefined
+
+            ColumnLayout {
+                spacing: 10
+                anchors.fill: parent
+                anchors.margins: 20
+
+                StyledText {
+                    anchors.left: parent.left
+                    text: i18.n + qsTr("List of duplicates:")
```
Ribtoks: These are more _"possible duplicates"_ than _"list of duplicates"_. I don't know the best wording, but I will leave it (comment) here to remember to fix this.

src/xpiks-qt/Dialogs/RemoveDuplicatesDialog.qml
```diff
@@ -0,0 +1,227 @@
...
+                                delegate: KeywordWrapper {
+                                    id: kw
+                                    isHighlighted: true
+                                    keywordText: keyword
+                                    hasSpellCheckError: !iscorrect
+                                    hasDuplicate: hasduplicate
+                                    itemHeight: flv.keywordHeight
+
+                                    onActionDoubleClicked: {
+                                        var flvModel = flv
```
Ribtoks: Um, what is this? This dialog should not allow to edit keywords.

Where did you get it from? Does the Suggestion dialog has this code?? (needs to be removed then)

src/xpiks-qt/Models/artitemsmodel.cpp
```diff
@@ -659,6 +659,16 @@ namespace Models {
 #endif
     }
 
+    void ArtItemsModel::retrieveDuplicates(int metadataIndex) {
+        LOG_INFO << "#";
```
Ribtoks: Dumb "#" should be `LOG_DEBUG` rather than `LOG_INFO`

src/xpiks-qt/Models/artitemsmodel.cpp
```diff
@@ -659,6 +659,16 @@ namespace Models {
 #endif
     }
 
+    void ArtItemsModel::retrieveDuplicates(int metadataIndex) {
```
Ribtoks: Why method called **retrieve** Duplicates returns `void`? I bet something like "setupDuplicatesModel" will work better here.

src/xpiks-qt/Models/artitemsmodel.cpp
```diff
@@ -687,6 +697,16 @@ namespace Models {
         }
     }
 
+    bool ArtItemsModel::hasSynonyms(int metadataIndex, int keywordIndex) const
+    {
```
Ribtoks: Braces on the same line, please

src/xpiks-qt/Models/artworkproxybase.cpp
```diff
@@ -203,6 +203,13 @@ namespace Models {
         m_CommandManager->setupSpellCheckSuggestions(metadataOperator, -1, Common::SuggestionFlags::All);
     }
 
+    void ArtworkProxyBase::doRetreiveDuplicates() {
```
Ribtoks: An English typo. [Retr I E ve] duplicates instead of [Retr E I ve] duplicates

src/xpiks-qt/Models/artworkproxybase.cpp
```diff
@@ -203,6 +203,13 @@ namespace Models {
...
+        LOG_DEBUG << "#";
+        auto *metadata = getBasicMetadataModel();
+        if (metadata)
```
Ribtoks: Please wrap everything with braces (`if` statement here)

src/xpiks-qt/Models/duplicatesmodel.cpp
```diff
@@ -0,0 +1,102 @@
+#include "duplicatesmodel.h"
+#include "../Commands/commandmanager.h"
+#include <QQmlEngine>
+
+namespace Models {
+
+namespace {
+    template<typename T, typename... Args>
```
Ribtoks: 1. Please leave some comments around why this is needed here when there is http://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique standard `std::make_unique`
2. If this is needed, than please name it differently to the standard one (even `my_make_unique` will work if you really need it)

src/xpiks-qt/Models/duplicatesmodel.cpp
```diff
@@ -0,0 +1,102 @@
...
+        m_KeywordsModel(make_unique<Common::BasicKeywordsModel>(m_Hold, this)) {
+        setSourceModel(m_KeywordsModel.get());
+        QObject::connect(this, SIGNAL(aboutToBeRemoved()), m_KeywordsModel.get(), SIGNAL(aboutToBeRemoved()));
+        QObject::connect(m_KeywordsModel.get(), SIGNAL(dataChanged(const QModelIndex &, const QModelIndex &, const QVector<int> &)), this, SIGNAL(invalidateFilter()));
+        QObject::connect(m_KeywordsModel.get(), SIGNAL(dataChanged(const QModelIndex &, const QModelIndex &, const QVector<int> &)), this, SLOT(onDataChanged()));
+    }
+
+    void DuplicatesModel::setupModel(Common::BasicKeywordsModel * model) {
+        beginResetModel();
+        LOG_INFO << '#';
```
Ribtoks: Please use `LOG_DEBUG` for dumb "#" and put it in the method header, not inside begin/end Reset

src/xpiks-qt/Models/duplicatesmodel.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+    int DuplicatesModel::getOriginalIndex(int index) const {
+        QModelIndex originalIndex = mapToSource(this->index(index, 0));
+        int row = originalIndex.row();
+
+        return row;
+    }
+
+    QObject *DuplicatesModel::getBasicModel()
+    {
```
Ribtoks: Braces on the same lines please..

src/xpiks-qt/Models/duplicatesmodel.cpp
```diff
@@ -0,0 +1,102 @@
...
+        m_KeywordsModel->removeKeywordAt(originalIndex, removed);
+    }
+
+    void DuplicatesModel::clearModel() {
+        LOG_DEBUG << "#";
+        m_KeywordsModel->clearKeywords();
+    }
+
+    void DuplicatesModel::saveChanges()
+    {
```
Ribtoks: Would be also nice to log_debug entries to this method

src/xpiks-qt/Models/duplicatesmodel.cpp
```diff
@@ -0,0 +1,102 @@
...
+        return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
+}
+
+}
+    DuplicatesModel::DuplicatesModel(QObject *parent):
+        QSortFilterProxyModel(parent),
+        Common::BaseEntity(),
+        m_ExternalModel(nullptr),
+        m_KeywordsModel(make_unique<Common::BasicKeywordsModel>(m_Hold, this)) {
+        setSourceModel(m_KeywordsModel.get());
```
Ribtoks: You have too many `m_KeywordsModel.get()` call here. Just do it once please.

src/xpiks-qt/Models/duplicatesmodel.cpp
```diff
@@ -0,0 +1,102 @@
...
+        QObject::connect(this, SIGNAL(aboutToBeRemoved()), m_KeywordsModel.get(), SIGNAL(aboutToBeRemoved()));
+        QObject::connect(m_KeywordsModel.get(), SIGNAL(dataChanged(const QModelIndex &, const QModelIndex &, const QVector<int> &)), this, SIGNAL(invalidateFilter()));
```
Ribtoks: I've replaced 400 calls to `QObject::connect` in master in order to have compile-time checks (this `connect()` you usedworks on strings and if it fails - you will know only in runtime). Maybe you can refactor this too to use method references 
like `QObject::connect(keywordsModel, &BasicKeywordsModel::dataChanged, this, &DuplicatesModel::dataChangedHandler)`. Maybe you will need to tweak method declarations to have all the required parameters.

src/xpiks-qt/Models/duplicatesmodel.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+        if (sourceRow < 0 || sourceRow >= size) {
+            return false;
+        }
+        result =  m_KeywordsModel->hasSynonyms(sourceRow);
+
+        return result;
+    }
+
+    bool DuplicatesModel::lessThan(const QModelIndex &sourceLeft, const QModelIndex &sourceRight) const {
```
Ribtoks: Why do you need sorting alltogether? I don't mind just having a heap of synonyms..

src/xpiks-qt/Models/duplicatesmodel.h
```diff
@@ -0,0 +1,40 @@
...
+    void onDataChanged();
+
+signals:
+        void aboutToBeRemoved();
+protected:
+        bool filterAcceptsRow(int sourceRow, const QModelIndex &sourceParent) const override;
+        bool lessThan(const QModelIndex &sourceLeft, const QModelIndex &sourceRight) const override;
+private:
+    Common::Hold m_Hold;
+    Common::BasicKeywordsModel * m_ExternalModel;
```
Ribtoks: I don't know, but it might be better to have a local _full_ copy of **one** basic keywords model instead of having all these pointers..

src/xpiks-qt/SpellCheck/spellcheckerservice.cpp
```diff
@@ -267,7 +267,8 @@ namespace SpellCheck {
 
     void SpellCheckerService::clearUserDictionary() {
         LOG_DEBUG << "#";
-        m_SpellCheckWorker->submitItem(std::shared_ptr<ISpellCheckItem>(new ModifyUserDictItem(true)));
+        if (m_SpellCheckWorker != nullptr)
```
Ribtoks: Please use braces for all sort of blocks.

src/xpiks-qt/SpellCheck/spellcheckitem.h
```diff
@@ -72,13 +76,19 @@ namespace SpellCheck {
...
             QObject(),
-            m_NeedsSuggestions(false) {}
+            m_NeedsSuggestions(false),
+            m_WordAnalysisFlag(wordAnalysisFlag)
+            {}
 
     public:
         const std::vector<std::shared_ptr<SpellCheckQueryItem> > &getQueries() const { return m_QueryItems; }
-        const QHash<QString, bool> &getHash() const { return m_SpellCheckResults; }
+        Common::WordAnalysisFlag getWordAnalysisFlag() const { return m_WordAnalysisFlag; }
```
Ribtoks: You used flag_S everywhere, but here you use just _flag_. Let's use plural everywhere for consistency.

src/xpiks-qt/SpellCheck/spellcheckworker.cpp
```diff
@@ -150,13 +150,18 @@ namespace SpellCheck {
...
+        const auto wordAnalysisFlag = item->getWordAnalysisFlag();
         bool anyWrong = false;
 
         if (!neededSuggestions) {
             size_t size = queryItems.size();
             for (size_t i = 0; i < size; ++i) {
                 auto &queryItem = queryItems.at(i);
-                bool isOk = checkWordSpelling(queryItem);
+                bool isOk = true;
+                if (HasFlag(wordAnalysisFlag, Common::WordAnalysisFlag::Spelling))
```
Ribtoks: Let's use braces for all sorts of blocks here. And 1 empty line after each block as well

src/xpiks-qt/SpellCheck/spellcheckworker.cpp
```diff
@@ -270,6 +275,27 @@ namespace SpellCheck {
...
 
+    QString SpellCheckWorker::getWordStem(const QString &word) {
+        QString result;
+
+        if (word.size() == 0) {
+            return result;
+        }
+        std::string encodedWord = m_Codec->fromUnicode(word).toStdString();
+        const std::vector<std::string> stems = m_Hunspell->stem(encodedWord);
+        if (!stems.empty())
```
Ribtoks: Block braces...

src/xpiks-qt/SpellCheck/spellcheckworker.cpp
```diff
@@ -270,6 +275,27 @@ namespace SpellCheck {
...
+            result =  QString::fromStdString(*std::min_element(stems.begin(), stems.end()));
```
Ribtoks: Let's not use function calls inside function calls. Better use a separate variable - will help for debugging.

src/xpiks-qt/Common/wordanalysisresult.h
```diff
@@ -0,0 +1,21 @@
+#include <QStringList>
```
RostaTasha: Done

src/xpiks-qt/Commands/commandmanager.cpp
```diff
@@ -296,6 +297,13 @@ void Commands::CommandManager::InjectDependency(Maintenance::MaintenanceService
     Q_ASSERT(maintenanceService != NULL); m_MaintenanceService = maintenanceService;
 }
 
+void Commands::CommandManager::InjectDependency(Models::DuplicatesModel *duplicatesModel)
+{
+    Q_ASSERT(duplicatesModel != NULL);
```
RostaTasha: Done

src/xpiks-qt/Common/basickeywordsmodel.h
```diff
@@ -57,12 +60,13 @@ namespace Common {
...
             KeywordRole = Qt::UserRole + 1,
-            IsCorrectRole
+            IsCorrectRole,
+            HasDuplicateRole
         };
 
     public:
 #ifdef CORE_TESTS
-        QVector<bool> &getSpellCheckResults() { return m_SpellCheckResults; }
+        QVector<KeywordSpellInfo> &getSpellCheckResults() { return m_SpellCheckResults; }
```
RostaTasha: Done

src/xpiks-qt/Common/basickeywordsmodel.h
RostaTasha: I don't understand.. it adds public if CORE_TESTS is defined

src/xpiks-qt/Common/basickeywordsmodel.h
```diff
@@ -168,6 +180,14 @@ namespace Common {
         bool isReplacedADuplicateUnsafe(int index, const QString &existingPrev,
                                         const QString &replacement) const;
         void emitSpellCheckChanged(int index=-1);
+#ifdef CORE_TESTS
+   public:
+#endif
+        void reevaluateStemsData();
+        void reevaluateStemsDataUnsafe();
```
RostaTasha: It is only for unit tests...but I have changed it.

src/xpiks-qt/Common/basicmetadatamodel.cpp
RostaTasha: Done

src/xpiks-qt/Common/basicmetadatamodel.cpp
RostaTasha: Done

src/xpiks-qt/Common/defines.h
RostaTasha: The integration test tests connectivity between components in terms of this feature. I wanted to exclude this factor in this test.

src/xpiks-qt/Common/wordanalysisresult.h
RostaTasha: Done

src/xpiks-qt/Dialogs/RemoveDuplicatesDialog.qml
RostaTasha: Done

src/xpiks-qt/Dialogs/RemoveDuplicatesDialog.qml
RostaTasha: Changed to possible duplicates

src/xpiks-qt/Dialogs/RemoveDuplicatesDialog.qml
RostaTasha: As discussed, I decided, that for long words, containing duplicates, like "girls" and "pretty young girl", the user would like not to delete keywords, but instead change them. Are you sure, that I should remove it? Please, confirm.

src/xpiks-qt/Models/artitemsmodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/artitemsmodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/artitemsmodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/artworkproxybase.cpp
RostaTasha: Changed to doSetupDuplicatesModel

src/xpiks-qt/Models/artworkproxybase.cpp
RostaTasha: Done

src/xpiks-qt/Models/duplicatesmodel.cpp
RostaTasha: Changed to my_make_unique

src/xpiks-qt/Models/duplicatesmodel.cpp
RostaTasha: It just extracts raw pointer from unique ptr, but I changed.

src/xpiks-qt/Models/duplicatesmodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/duplicatesmodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/duplicatesmodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/duplicatesmodel.cpp
RostaTasha: Mmm. I thought that it makes life easier. Please, conform, that it should be removed.

src/xpiks-qt/Models/duplicatesmodel.h
RostaTasha: I needanyway pointer to m_ExternalModel, because it is the destination, where I will write to in case of save.

Choice of unique_pointer is dictated by these const limitations of inherited methods QAbstractProxyModel and requirement of usage of mutexes at some level..

src/xpiks-qt/SpellCheck/spellcheckerservice.cpp
RostaTasha: Done

src/xpiks-qt/SpellCheck/spellcheckitem.h
RostaTasha: Done

src/xpiks-qt/SpellCheck/spellcheckworker.cpp
RostaTasha: Done

src/xpiks-qt/SpellCheck/spellcheckworker.cpp
RostaTasha: Done

src/xpiks-qt/SpellCheck/spellcheckworker.cpp
RostaTasha: Done

src/xpiks-qt/Models/duplicatesmodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/duplicatesmodel.h
Ribtoks: Then you should work with `Common::IMetadataOperator` so it will be possible to use your function both from the main grid and from the dialogs - One item editing + Multiple items editing.

BTW, did you add this functionality there as well? If not, it should be there..

src/xpiks-qt/Models/duplicatesmodel.cpp
Ribtoks: Yes, I can confirm sorting and editing shall not be used there. Let's have a set of tools which do one job at a time.

src/xpiks-qt/Commands/commandmanager.cpp
```diff
@@ -715,14 +723,15 @@ void Commands::CommandManager::generatePreviews(const QVector<Models::ArtworkMet
 
 void Commands::CommandManager::submitKeywordForSpellCheck(Common::BasicKeywordsModel *item, int keywordIndex) const {
     Q_ASSERT(item != NULL);
-    if ((m_SettingsModel != NULL) && m_SettingsModel->getUseSpellCheck() && (m_SpellCheckerService != NULL)) {
+    Common::WordAnalysisFlag wordanalysisFlag = getSpellCheckFlag();
+    if ((wordanalysisFlag != Common::WordAnalysisFlag::None) && (m_SpellCheckerService != NULL)) {
```
RostaTasha: Renamed getSpellCheckFlag() to getWordAnalysisFlag ()

src/xpiks-qt/Common/wordanalysisresult.h
RostaTasha: Tried to fix..

src/xpiks-qt/Models/duplicatesmodel.cpp
RostaTasha: Ok..

src/xpiks-qt/Common/flags.h
```diff
@@ -170,6 +170,13 @@ namespace Common {
         Spelling = 4
     };
 
+    enum struct WordAnalysisFlag {
```
Ribtoks: Please use form:
`enum struct WordAnalysisFlag: Common::flag_t { ... }`
So all flags will have an underlying type `uint32_t` (typdeffed as `Common::flag_t`

src/xpiks-qt/Commands/commandmanager.cpp
```diff
@@ -1072,3 +1092,19 @@ void Commands::CommandManager::clearCurrentItem() const {
         m_UIManager->clearCurrentItem();
     }
 }
+
+Common::WordAnalysisFlag Commands::CommandManager::getWordAnalysisFlag() const
```
Ribtoks: Please fix up all the braces style in this method accordingly to typical braces style.

src/xpiks-qt/Common/basickeywordsmodel.h
```diff
@@ -168,6 +182,14 @@ namespace Common {
         bool isReplacedADuplicateUnsafe(int index, const QString &existingPrev,
                                         const QString &replacement) const;
         void emitSpellCheckChanged(int index=-1);
+        void reevaluateStemsDataUnsafe();
```
Ribtoks: Please add a visibility specifier like `public:`. Otherwise methods starting from this comment will belong to `signals:`

src/xpiks-qt/Dialogs/RemoveDuplicatesDialog.qml
```diff
@@ -0,0 +1,264 @@
...
+                    }
+                }
+
+                Rectangle {
+                    Layout.fillWidth: true
+                    Layout.fillHeight: true
+                    color: Colors.defaultControlColor
+
+                    Item {
+                        visible: duplicatesListView.count == 0
```
Ribtoks: I guess we shall also assert somewhere that we don't show a dialog if we have `0` "suggestions" (should also be duplicates in this context)

src/xpiks-qt/Dialogs/RemoveDuplicatesDialog.qml
```diff
@@ -0,0 +1,264 @@
...
+                            id: duplicatesListView
+                            model: duplicatesModel
+                            focus: true
+                            spacing: 5
+
+                            delegate: Rectangle {
+                                id: duplicatesWrapper
+                                property int delegateIndex: index
+                                color: Colors.panelColor
+                                width: parent.width - 10
```
Ribtoks: Is this `width-10` code copied from Suggestion Popup? I guess in this case we can also use `anchors.left: parent.left` and anchor margin (+ same stuff for "right") for better code

src/xpiks-qt/Dialogs/SettingsWindow.qml
```diff
@@ -404,6 +404,23 @@ ApplicationWindow {
                             }
                         }
 
+                        StyledCheckbox {
+                            id: autoDuplicateSearchCheckbox
+                            text: i18.n + qsTr("Check duplicates automatically")
```
Ribtoks: I would say its more like "Detect duplicates", not "Check duplicates" (from the English language point of view)

Also might be useful to have it (_Detect Duplicates_) instead of _Use Duplicates Check_

src/xpiks-qt/Models/artitemsmodel.cpp
```diff
@@ -659,6 +659,16 @@ namespace Models {
 #endif
     }
 
+    void ArtItemsModel::setupDuplicatesModel(int metadataIndex) {
+        LOG_DEBUG << "#";
```
Ribtoks: I would say its better to `LOG_INFO << metadataIndex;` in this case so you would know from the logs which index have been passed if the method will not do any work.

src/xpiks-qt/Models/artitemsmodel.cpp
```diff
@@ -687,6 +697,15 @@ namespace Models {
         }
     }
 
+    bool ArtItemsModel::hasSynonyms(int metadataIndex, int keywordIndex) const {
+        if (0 <= metadataIndex && metadataIndex < getArtworksCount()) {
+            ArtworkMetadata *metadata = m_ArtworkList.at(metadataIndex);
```
Ribtoks: You check if metadata is not `NULL` in method `ArtItemsModel::setupDuplicatesModel()` and don't check here. Its better to be consistent and either check or not check (then `assert`) in both methods.

src/xpiks-qt/Models/duplicatesmodel.cpp
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    QStringList SimpleKeywordsModel::getSelected() const {
+        QStringList result;
+        size_t size = m_Keywords.size();
+
+        for (int i = 0; i < size; i++) {
+            bool selected = m_Keywords[i].second;
+            if (selected) {
+                result.append(m_Keywords[i].first);
```
Ribtoks: I know it does not matter a lot here, but you don't reserve space in a `QStringList`. It should be a habit.

src/xpiks-qt/Models/duplicatesmodel.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+namespace Models {
+    namespace {
+        template<typename T, typename ... Args>
+        std::unique_ptr<T> my_make_unique(Args && ... args) {
+            return std::unique_ptr<T>(new T(std::forward<Args>(args) ...));
+        }
+    }
+
+    SimpleKeywordsModel::SimpleKeywordsModel(const QString &stem, const QStringList &keywords):
```
Ribtoks: I wouldn't call it `Simple KeywordsModel`. First of all, because usually people add stuff to classes over time and this model will not be "simple" anymore. Also it does not reflect what does this model actually do (select keywords). So I would go for `SelectableKeywordsModel` or anything which better reflects what it does.

src/xpiks-qt/Models/duplicatesmodel.cpp
```diff
@@ -0,0 +1,208 @@
...
+            case KeywordRole:
+                return item.first;
+            case IsSelectedRole:
+                return item.second;
+            default:
+                return QVariant();
+        }
+    }
+
+    bool SimpleKeywordsModel::setData(const QModelIndex &index, const QVariant &value, int role) {
```
Ribtoks: I'm not sure this `setData()` is needed at all.
1. For `setData()` and Qt compatibility you should also provide method `editableFlags()` (or smth like that - I don't remember the name - should be in the ArtItemsModel)
2. You can just have a methods `selectKeyword(int)` which will do same stuff faster (`setData()` involves some additional Qt/QML constructs).

src/xpiks-qt/Models/duplicatesmodel.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+        return result;
+    }
+
+    void SimpleKeywordsModel::onSelectedChanged(const QString &keyword, bool value) {
+        LOG_DEBUG << '#';
+        size_t size = m_Keywords.size();
+        for (int i = 0; i < size; i++) {
+            const auto &itemKeyword = m_Keywords[i].first;
+            if (itemKeyword == keyword) {
```
Ribtoks: 1. There shouldn't be any real clones of the keywords in the list. Why don't you `break` from the loop?
2. Why do you need this method? It should only change from private methods of this model, not from outside. The model should select/deselect All/One item and that's it - no bindings to anything external. Feel free to correct me if I'm wrong.

src/xpiks-qt/Models/duplicatesmodel.cpp
```diff
@@ -0,0 +1,208 @@
...
+        m_Duplicates.clear();
+        const auto stemsWithDuplicates = model->getDuplicatesModel();
+        const auto begin = stemsWithDuplicates.begin();
+        const auto end = stemsWithDuplicates.end();
+        for (auto it = begin; it != end; it++) {
+            const auto &stem = it.key();
+            const auto &duplicates = it.value();
+            m_Duplicates.emplace_back(new SimpleKeywordsModel(stem, duplicates));
+            const auto *item = m_Duplicates.back().get();
+            QObject::connect(item, SIGNAL(selectedChanged(const QString &, bool)), this, SLOT(onSelectedChanged(const QString &, bool)));
```
Ribtoks: Please use form of `QObject::connect` with c++11 method pointers instead of strings comparison for compile-time checks.

src/xpiks-qt/Models/duplicatesmodel.cpp
```diff
@@ -0,0 +1,208 @@
...
+        }
+    }
+
+    void DuplicatesModel::onSelectedChanged(const QString &keyword, bool value) {
+        LOG_DEBUG << '#';
+        emit broadcastSelected(keyword, value);
+    }
+
+    QHash<int, QByteArray> DuplicatesModel::roleNames() const {
+        QHash<int, QByteArray> roles;
```
Ribtoks: I don't know if that makes any difference but I tend to fill `roles` from base classes' `roleNames()` first. For the compatibility. Something like asserting.

src/xpiks-qt/Models/duplicatesmodel.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+    void DuplicatesModel::saveChanges() {
+        QStringList keywordsToRemove;
+
+        for (const auto &simpleModel : m_Duplicates) {
+            keywordsToRemove.append(simpleModel->getSelected());
+        }
+
+        if (m_ExternalModel != nullptr) {
+            m_ExternalModel->removeKeywords(QSet<QString>::fromList(keywordsToRemove), true);
```
Ribtoks: Please don't use method calls in the parameters.

src/xpiks-qt/Models/duplicatesmodel.h
```diff
@@ -0,0 +1,95 @@
...
+
+namespace Models {
+    class SimpleKeywordsModel:
+        public QAbstractListModel
+    {
+    Q_OBJECT
+
+    public:
+        SimpleKeywordsModel(const QString &stem, const QStringList &keywords);
+        ~SimpleKeywordsModel() = default;
```
Ribtoks: This class is not trivial and derived - let's better have empty virtual destructor.

src/xpiks-qt/Models/duplicatesmodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/duplicatesmodel.cpp
RostaTasha: Changed implementation

src/xpiks-qt/Models/duplicatesmodel.cpp
RostaTasha: Changed implementation

src/xpiks-qt/Models/duplicatesmodel.cpp
RostaTasha: Changed implementation

src/xpiks-qt/Common/flags.h
RostaTasha: Done

src/xpiks-qt/Commands/commandmanager.cpp
RostaTasha: Done

src/xpiks-qt/Common/basickeywordsmodel.h
RostaTasha: Done

src/xpiks-qt/Dialogs/SettingsWindow.qml
RostaTasha: Done

src/xpiks-qt/Dialogs/RemoveDuplicatesDialog.qml
RostaTasha: Done

src/xpiks-qt/Dialogs/RemoveDuplicatesDialog.qml
RostaTasha: Done

src/xpiks-qt/Models/artitemsmodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/artitemsmodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/duplicatesmodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/duplicatesmodel.cpp
RostaTasha: Changed implementation

src/xpiks-qt/Models/duplicatesmodel.h
RostaTasha: It is not derived.

src/xpiks-qt/Models/duplicatesmodel.cpp
RostaTasha: I need both reading and writing of "selected" property.  I did not understand the alternative to using of data/setdata mechanism. Could you, please, clarify?

src/xpiks-qt/Commands/commandmanager.cpp
```diff
@@ -747,14 +755,15 @@ void Commands::CommandManager::generatePreviews(const QVector<Models::ArtworkMet
...
+    Common::WordAnalysisFlag wordanalysisFlag = getWordAnalysisFlag();
+    if ((wordanalysisFlag != Common::WordAnalysisFlag::None) && (m_SpellCheckerService != NULL)) {
         m_SpellCheckerService->submitKeyword(item, keywordIndex);
     }
 }
 
 void Commands::CommandManager::submitForSpellCheck(const QVector<Models::ArtworkMetadata *> &items) const {
-    if ((m_SettingsModel != NULL) &&
-        m_SettingsModel->getUseSpellCheck() &&
+    Common::WordAnalysisFlag wordanalysisFlag = getWordAnalysisFlag();
```
Ribtoks: I guess it's the copy-paste typo everywhere. Should be `wordAnalysisFlag` (camel case)

src/xpiks-qt/Commands/commandmanager.cpp
```diff
@@ -1201,3 +1224,18 @@ void Commands::CommandManager::clearCurrentItem() const {
         m_UIManager->clearCurrentItem();
     }
 }
+
+Common::WordAnalysisFlag Commands::CommandManager::getWordAnalysisFlag() const {
+    Common::WordAnalysisFlag result;
+    if (m_SettingsModel != NULL)
```
Ribtoks: Please keep up with the braces style of Xpiks

src/xpiks-qt/Commands/commandmanager.cpp
```diff
@@ -1201,3 +1224,18 @@ void Commands::CommandManager::clearCurrentItem() const {
         m_UIManager->clearCurrentItem();
     }
 }
+
+Common::WordAnalysisFlag Commands::CommandManager::getWordAnalysisFlag() const {
+    Common::WordAnalysisFlag result;
```
Ribtoks: Should be explicitly initialized to NONE

src/xpiks-qt/Common/basickeywordsmodel.h
Ribtoks: I cannot see where is it "Done". Please add it _above_ the `bool isReplacedADuplicateUnsafe()`

src/xpiks-qt/Common/basicmetadatamodel.cpp
```diff
@@ -183,6 +183,11 @@ namespace Common {
         return BasicKeywordsModel::appendPreset(presetList);
     }
 
+    QHash<QString, QStringList> BasicMetadataModel::getDuplicatesModel()
```
Ribtoks: Why it should return `QHash<>` by value and why it is then not-const?

src/xpiks-qt/Common/flags.h
```diff
@@ -172,6 +172,13 @@ namespace Common {
         Spelling = 4
     };
 
+    enum struct WordAnalysisFlag: flag_t {
```
Ribtoks: I suppose naming should be flagS, not flag since you have `All` which is a combination.

src/xpiks-qt/Common/wordanalysisresult.h
Ribtoks: Please try again

src/xpiks-qt/Common/wordanalysisresult.h
```diff
@@ -0,0 +1,62 @@
...
+    struct KeywordSpellInfo {
+        QStringList m_Stems;
+        bool m_IsCorrect;
+
+        KeywordSpellInfo():
+            m_Stems(), m_IsCorrect(true)
+        {}
+
+        void reset()
+        {
```
Ribtoks: Braces style please

src/xpiks-qt/Models/artitemsmodel.cpp
```diff
@@ -663,6 +663,15 @@ namespace Models {
 #endif
     }
 
+    void ArtItemsModel::setupDuplicatesModel(int metadataIndex) {
+        LOG_DEBUG << metadataIndex;
```
Ribtoks: Should be `LOG_INFO`

src/xpiks-qt/Models/artitemsmodel.cpp
```diff
@@ -691,6 +700,16 @@ namespace Models {
         }
     }
 
+    bool ArtItemsModel::hasSynonyms(int metadataIndex, int keywordIndex) const {
```
Ribtoks: It is everywhere `hasDuplicates()` but is `hasSynonyms()` here. Is it by design or just a typo?

src/xpiks-qt/Models/artworkmetadata.cpp
```diff
@@ -107,6 +107,16 @@ namespace Models {
         return isInDir;
     }
 
+    bool ArtworkMetadata::hasSynonyms(int keywordIndex) const
```
Ribtoks: 1. Braces
2. It would be nice to have some sort of comment explaining the difference between `synonyms` and `duplicates` because I don't understand it right now.

src/xpiks-qt/Models/artworkmetadata.h
```diff
@@ -150,6 +151,7 @@ namespace Models {
         virtual int appendKeywords(const QStringList &keywordsList) override;
         virtual bool removeKeywords(const QSet<QString> &keywordsSet, bool caseSensitive=true) override;
         virtual QString getKeywordsString() override { return m_MetadataModel.getKeywordsString(); }
+        virtual QHash<QString, QStringList> getDuplicatesModel() override { return m_MetadataModel.getDuplicatesModel(); }
```
Ribtoks: Should it be copy-by-value here or it's a typo?

src/xpiks-qt/Models/duplicatesmodel.cpp
```diff
@@ -0,0 +1,229 @@
...
+                result = true;
+                break;
+            default:
+                return false;
+        }
+
+        return result;
+    }
+
+    Qt::ItemFlags SelectableKeywordsModel::flags(const QModelIndex &index) const
```
Ribtoks: Please keep up with the braces style

src/xpiks-qt/Models/duplicatesmodel.cpp
```diff
@@ -0,0 +1,229 @@
...
+        switch (role) {
+            case IsSelectedRole:
+                if (m_DuplicatesModel->setSelected(keywordIndex, value.toBool())) {
+                    emit dataChanged(index, index, QVector<int>() << IsSelectedRole);
+                }
+
+                result = true;
+                break;
+            default:
+                return false;
```
Ribtoks: Why to put this `return` here if the default `result` is still false. This `default` clause can only have `break` or smth like that

src/xpiks-qt/Models/duplicatesmodel.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+        const auto begin = stemsWithDuplicates.begin();
+        const auto end = stemsWithDuplicates.end();
+        for (auto it = begin; it != end; it++) {
+            const auto &stem = it.key();
+            const auto &duplicates = it.value();
+            QVector<int> indexes;
+            indexes.reserve(duplicates.size());
+            for (const auto &duplicate : duplicates) {
+                indexes.push_back(duplicatesList.indexOf(duplicate));
```
Ribtoks: what if `indexOf()` returns -1?

Also, please don't put function calls inside parameters

src/xpiks-qt/Models/duplicatesmodel.cpp
```diff
@@ -0,0 +1,229 @@
...
+        }
+
+        return QAbstractItemModel::flags(index) | Qt::ItemIsEditable;
+    }
+
+    bool DuplicatesModel::setSelected(int index, bool value) {
+        int size = m_KeywordsData.size();
+
+        if (0 <= index && index < size) {
+            m_KeywordsDataLock.lockForWrite();
```
Ribtoks: Please use `QWriteLocker` for this intent to make use of C++ RAII magic.

src/xpiks-qt/Models/duplicatesmodel.h
```diff
@@ -0,0 +1,89 @@
...
+    class DuplicatesModel;
+
+    class SelectableKeywordsModel:
+        public QAbstractListModel
+    {
+    Q_OBJECT
+
+    public:
+        SelectableKeywordsModel(const QString &stem, const QVector<int> &keywordsIndexes, DuplicatesModel *duplicatesModel);
+        ~SelectableKeywordsModel() = default;
```
Ribtoks: Please make this `virtual`

src/xpiks-qt/Models/duplicatesmodel.h
```diff
@@ -0,0 +1,89 @@
...
+// QAbstractListModel
+
+    protected:
+        enum Roles {
+            KeywordRole = Qt::UserRole + 1,
+            IsSelectedRole
+        };
+
+    public:
+        QString getStem() const { return m_Stem; }
```
Ribtoks: Why it is returned by value, not by reference?

src/xpiks-qt/SpellCheck/spellcheckerservice.cpp
```diff
@@ -290,4 +292,20 @@ namespace SpellCheck {
         LOG_INFO << "Size of dictionary:" << number << "word(s)";
         emit userDictWordsNumberChanged();
     }
+
+    Common::WordAnalysisFlag SpellCheckerService::getSpellCheckFlag() const
+    {
+        Common::WordAnalysisFlag result;
```
Ribtoks: Please explicitly initialize that to NONE

src/xpiks-qt/SpellCheck/spellcheckworker.cpp
```diff
@@ -150,13 +150,22 @@ namespace SpellCheck {
...
             for (size_t i = 0; i < size; ++i) {
                 auto &queryItem = queryItems.at(i);
-                bool isOk = checkWordSpelling(queryItem);
+                bool isOk = true;
+                if (HasFlag(wordAnalysisFlags, Common::WordAnalysisFlag::Spelling)) {
+                    isOk = checkWordSpelling(queryItem);
+                }
+
+                if (HasFlag(wordAnalysisFlags, Common::WordAnalysisFlag::Stemming)) {
+                    getWordStem(queryItem);
```
Ribtoks: What is being done with the results of method `getWordStem()`? Why it is ignored? If it should be ignored, why this method returns anything in the first place?

src/xpiks-qt/SpellCheck/spellcheckworker.cpp
```diff
@@ -270,6 +279,29 @@ namespace SpellCheck {
...
+        QString result;
+
+        if (word.size() == 0) {
+            return result;
+        }
+        std::string encodedWord = m_Codec->fromUnicode(word).toStdString();
+        const std::vector<std::string> stems = m_Hunspell->stem(encodedWord);
+        bool stemsNotEmpty = !stems.empty();
+        if (stemsNotEmpty) {
+            result =  QString::fromStdString(*std::min_element(stems.begin(), stems.end()));
```
Ribtoks: Please don't put function calls into parameters.

src/xpiks-tests/xpiks-tests-integration/main.cpp
```diff
@@ -567,6 +568,7 @@ int main(int argc, char *argv[]) {
     integrationTests.append(new UserDictEditTest(&commandManager));
     integrationTests.append(new WeirdNamesReadTest(&commandManager));
     integrationTests.append(new RestoreSessionTest(&commandManager));
+    integrationTests.append(new DuplicateSearch(&commandManager));
```
Ribtoks: Why this test does not end with word "Test" like _every_ other integration test? Please fix the name.

RostaTasha: Done

src/xpiks-qt/SpellCheck/spellcheckworker.cpp
RostaTasha: Done

src/xpiks-qt/SpellCheck/spellcheckworker.cpp
RostaTasha: Refactored

src/xpiks-qt/SpellCheck/spellcheckerservice.cpp
RostaTasha: Done

src/xpiks-qt/Models/duplicatesmodel.h
RostaTasha: Done

src/xpiks-qt/Models/duplicatesmodel.h
RostaTasha: Removed explicit destructor declaration

src/xpiks-qt/Models/duplicatesmodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/duplicatesmodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/duplicatesmodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/duplicatesmodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/artitemsmodel.cpp
RostaTasha: Done

src/xpiks-qt/Common/wordanalysisresult.h
RostaTasha: Done

src/xpiks-qt/Commands/commandmanager.cpp
RostaTasha: Done

src/xpiks-qt/Commands/commandmanager.cpp
RostaTasha: Done

src/xpiks-qt/Commands/commandmanager.cpp
RostaTasha: Done

src/xpiks-qt/Common/basicmetadatamodel.cpp
RostaTasha: Done

src/xpiks-qt/Common/flags.h
RostaTasha: Done

src/xpiks-qt/Models/artitemsmodel.cpp
RostaTasha: Done

src/xpiks-qt/Models/artworkmetadata.cpp
RostaTasha: Refactored

src/xpiks-qt/Models/artworkmetadata.h
RostaTasha: Changed

Ribtoks:
Merged to master as another branch.

Issue #412, opened 2017-03-27
Option to save the session of opened files/directories


Ribtoks:
Save additional file _xpks_session_ or smth like that in .json format with few properties: 

- list of all opened files
- some flags / opened buffers?

In C++ create `SessionManager` to handle that.
Add menu [File -> Session Manager -> [x] Save session] path
Add checkbox to Settings to handle that as well.

gnidorah:
Working on this task

PR #439, opened 2017-06-09
Option to save the session of opened files/directories

gnidorah:
For #412
Thank you!


src/xpiks-qt/Maintenance/savesessionjobitem.cpp
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+#include "../Models/sessionmanager.h"
+#include "savesessionjobitem.h"
+
+namespace Maintenance {
+    SaveSessionJobItem::SaveSessionJobItem(std::unique_ptr<MetadataIO::SessionSnapshot> &sessionSnapshot, Models::SessionManager *sessionManager):
+        m_SessionSnapshot(std::move(sessionSnapshot)),
+        m_SessionManager(sessionManager)
+    {
```
Ribtoks: Would be nice to assert that `m_SessionManager` is not null

src/xpiks-qt/MetadataIO/metadataiocoordinator.cpp
```diff
@@ -214,15 +228,23 @@ namespace MetadataIO {
         }
     }
 
-    void MetadataIOCoordinator::continueWithoutReading() {
+    void MetadataIOCoordinator::continueWithoutReading(Models::ArtworkMetadata* restoredMetadata) {
```
Ribtoks: Raw pointers and all this stuff does not look good. Let's talk in chat more.

src/xpiks-qt/MetadataIO/metadataiocoordinator.cpp
```diff
@@ -273,12 +303,23 @@ namespace MetadataIO {
...
+                metadata->setBackupIgnored(ignoreBackups);
             }
         }
 
-        afterImportHandler(itemsToRead, ignoreBackups);
+        afterImportHandler(*itemsToRead, ignoreBackups);
 
-        emit metadataReadingFinished();
+        if (restoredMetadata) {
+            Warnings::WarningsModel *warningsModel = m_CommandManager->getWarningsModel();
```
Ribtoks: Why do you need this? `WarningsModel` is just a proxy model over `ArtItemsModel` which means if you add/remove items to `ArtItemsModel`, then `WarningsModel` will automagically be updated.

src/xpiks-qt/MetadataIO/artworkmetadatasnapshot.h
```diff
@@ -0,0 +1,69 @@
...
+        ArtworkMetadataSnapshot();
+
+        ArtworkMetadataSnapshot(Models::ArtworkMetadata *metadata);
+        virtual ~ArtworkMetadataSnapshot();
+    };
+
+    template <typename T>
+    class ArtworksSnapshotBase {
+    public:
+        ArtworksSnapshotBase(T *artworksList):
```
gnidorah: Sorry, It doesn't work. If I switch from std::vector<MetadataIO::ArtworkMetadataSnapshot *> and *snapshot = new MetadataIO::ArtworkMetadataSnapshot(artwork) to std::vector<MetadataIO::ArtworkMetadataSnapshot> and MetadataIO::ArtworkMetadataSnapshot snapshot(artwork) then I get assert in Common::Hold::acquire() {Q_ASSERT(prev > 0);}.

src/xpiks-qt/Models/artworkmetadata.h
```diff
@@ -60,18 +60,27 @@ namespace Models {
             FlagIsModified = 1 << 0,
             FlagsIsSelected = 1 << 1,
             FlagIsInitialized = 1 << 2,
-            FlagIsUnavailable = 1 << 3
+            FlagIsUnavailable = 1 << 3,
+            FlagIsMetadataImported = 1 << 4,
+            FlagIsBackupIgnored = 1 << 5,
+            FlagIsRestored = 1 << 6
```
gnidorah: It's not finished: RemoveArtworksCommand::execute() and UndoRedo::RemoveArtworksHistoryItem::undo() should operate an info on whether file is restored and if yes it's metadata state, to work equally to session restoration on Xpiks launch.
But I think I don't get you. If there will be no difference, then user will get metadata import dialog on session restoration, like if files were added manually. Is it what is needed? No info about metadata import preference per file?

src/xpiks-qt/MetadataIO/metadataiocoordinator.cpp
```diff
@@ -87,13 +88,26 @@ namespace MetadataIO {
         }
 
         m_IsImportInProgress = false;
+
+        for (auto item: itemsToRead) {
```
gnidorah: I've optimized this, but again, with per-file metadata import prefrence any combination of continueReading(ignoreBackups = true), continueReading(ignoreBackups = false) and continueWithoutReading() may be called at once.

src/xpiks-qt/MetadataIO/metadataiocoordinator.cpp
gnidorah: If I remove it, "Check warnings" label in the left bottom corner doesn't become orange if there are warnings, unless I click on it.
This is also what Import Metadata dialog does: https://github.com/Ribtoks/xpiks/blob/master/src/xpiks-qt/Dialogs/ImportMetadata.qml#L236

src/xpiks-qt/Models/artworkmetadata.h
```diff
@@ -221,6 +236,29 @@ namespace Models {
...
+    {
+    public:
+        ArtworkMetadataLocker(Models::ArtworkMetadata *metadata):
+            m_ArtworkMetadata(metadata)
+        {
+            if (m_ArtworkMetadata != nullptr) {
+                m_ArtworkMetadata->acquire();
+            }
+        }
+        ~ArtworkMetadataLocker() {
```
Ribtoks: This class is inherited so it's destructor should be declared `virtual`

src/xpiks-qt/Models/sessionmanager.cpp
```diff
@@ -0,0 +1,166 @@
...
+            if (filePath.isEmpty()) {
+                continue;
+            }
+
+            qint64 directoryID = 0;
+            if (artworksRepository->accountFile(filePath, directoryID)) {
+                Models::ArtworkMetadata *metadata = artItemsModel->createMetadata(filePath, directoryID);
+                metadata->setRestored();
+
+                LOG_INTEGRATION_TESTS << "Added file for session restore:" << filePath;
```
Ribtoks: 1. Why logging only for integration tests?
2. "Added file **from** session restore". Not "for" ?

src/xpiks-qt/Commands/addartworkscommand.h
```diff
@@ -24,17 +24,23 @@
...
+namespace Models {
+    class ArtworkMetadata;
+}
+
 namespace Commands {
     class AddArtworksCommand : public CommandBase
     {
     public:
-        AddArtworksCommand(const QStringList &pathes, const QStringList &vectorPathes, bool autoDetectVectors) :
+        AddArtworksCommand(const QStringList &pathes, const QStringList &vectorPathes, bool autoDetectVectors, std::vector<Models::ArtworkMetadata *> *restoredMetadata = NULL) :
```
Ribtoks: Sounds like there's a need for another constructor for restored metadata and all other cases

src/xpiks-qt/MetadataIO/metadataiocoordinator.cpp
```diff
@@ -87,13 +88,40 @@ namespace MetadataIO {
...
+                    discarded.append(item);
+                }
+            }
+        }
+        if (!discarded.isEmpty()) {
+            continueWithoutReading(discarded);
+        }
+        if (!importedWithoutBackup.isEmpty()) {
+            m_CanProcessResults = true;
+            readingFinishedHandler(true, importedWithoutBackup);
```
Ribtoks: This handler can only be called once, but your code logic allows it to run twice which is undesirable at all.

src/xpiks-qt/MetadataIO/metadataiocoordinator.cpp
```diff
@@ -214,15 +242,21 @@ namespace MetadataIO {
...
 
-    void MetadataIOCoordinator::continueWithoutReading() {
+    void MetadataIOCoordinator::continueWithoutReading(QVector<Models::ArtworkMetadata*> restoredMetadata) {
         LOG_DEBUG << "Setting technical data";
 
         const QHash<QString, ImportDataResult> &importResult = m_ReadingWorker->getImportResult();
-        const QVector<Models::ArtworkMetadata*> &itemsToRead = m_ReadingWorker->getItemsToRead();
+        const QVector<Models::ArtworkMetadata*> *itemsToRead;
+
+        if (!restoredMetadata.isEmpty()) {
```
Ribtoks: This needs to be in a different method. You're mixing here two very different concepts and your code does not look pleasant to read because of it.

src/xpiks-qt/MetadataIO/artworkmetadatasnapshot.h
```diff
@@ -0,0 +1,67 @@
...
+    };
+
+    template <typename T>
+    class ArtworksSnapshotBase {
+    public:
+        ArtworksSnapshotBase(T &artworksList)
+        {
+            LOG_DEBUG << "#";
+
+            for (auto artwork: artworksList) {
```
Ribtoks: You need to `reserve()` space for your elements otherwise it will be slow on large artworks collections.

src/xpiks-qt/Models/sessionmanager.cpp
```diff
@@ -0,0 +1,166 @@
...
+                }
+
+                if (obj.contains(METADATAIMPORTED_KEY)) {
+                    metadata->setMetadataImported();
+                    if (obj.contains(BACKUPIGNORED_KEY)) {
+                        metadata->setBackupIgnored(true);
+                    }
+                }
+
+                m_MetadataVector.push_back(metadata);
```
Ribtoks: You have to `reserve()` space for `m_MetadataVector` otherwise it will be slow for large sessions.

src/xpiks-qt/Maintenance/maintenanceservice.cpp
```diff
@@ -128,6 +129,11 @@ namespace Maintenance {
         m_MaintenanceWorker->submitFirst(jobItem);
     }
 
+    void MaintenanceService::saveSession(std::unique_ptr<MetadataIO::SessionSnapshot> &sessionSnapshot, Models::SessionManager *sessionManager) {
+        std::shared_ptr<IMaintenanceItem> jobItem(new SaveSessionJobItem(sessionSnapshot, sessionManager));
```
Ribtoks: I guess you can also log debug here

src/xpiks-qt/MetadataIO/artworkmetadatasnapshot.h
```diff
@@ -0,0 +1,68 @@
...
+            Models::ArtworkMetadataLocker(metadata)
+        {
+        }
+    };
+
+    template <typename T>
+    class ArtworksSnapshotBase {
+    public:
+        ArtworksSnapshotBase(T &artworksList)
+        {
```
Ribtoks: Please keep up with the braces style

src/xpiks-qt/MetadataIO/artworkmetadatasnapshot.h
```diff
@@ -0,0 +1,68 @@
...
+            LOG_DEBUG << "#";
+
+            m_Snapshot.reserve(artworksList.size());
+            for (auto artwork: artworksList) {
+                auto *snapshot = new MetadataIO::ArtworkMetadataSnapshot(artwork);
+                m_Snapshot.push_back(snapshot);
+            }
+        }
+        ~ArtworksSnapshotBase() {
```
Ribtoks: If there's any meaningful destructor in a class, it should better be marked `virtual`. This is because somebody will inherit your class in future and forget to make base classes destructor virtual and will be possibly leaking memory. Only if the destructor is empty, you can omit it.

src/xpiks-qt/MetadataIO/artworkmetadatasnapshot.h
```diff
@@ -0,0 +1,68 @@
...
+            LOG_DEBUG << "#";
+
+            qDeleteAll(m_Snapshot);
+        }
+
+    public:
+        const std::vector<MetadataIO::ArtworkMetadataSnapshot *> &getSnapshot() { return m_Snapshot; }
```
Ribtoks: Should it be `const` method?

src/xpiks-qt/MetadataIO/metadataiocoordinator.cpp
```diff
@@ -92,8 +93,9 @@ namespace MetadataIO {
     void MetadataIOCoordinator::writingWorkerFinished(bool success) {
         LOG_INFO << success;
         setHasErrors(!success);
-        const QVector<Models::ArtworkMetadata*> &artworksToWrite = m_WritingWorker->getItemsToWrite();
-        m_CommandManager->addToLibrary(artworksToWrite);
+        std::vector<Models::ArtworkMetadata*> artworksToWrite = m_WritingWorker->getItemsToWrite().toStdVector();
```
Ribtoks: You're instantiating your custom classes in the `ArtworksSnapshotBase` from `std::vector`. With the same success you can create an overload where you instantiate them from `QVector`. But here you do double work - first you convert `QVector` to `std::vector` and then in the `ArtworksSnapshot` constructor you instantiate yet another vector from `std::vector`. Instead of A-B, B-C, you can go A-C directly.

src/xpiks-qt/Models/artworkmetadata.h
```diff
@@ -221,6 +221,29 @@ namespace Models {
...
+            }
+        }
+        virtual ~ArtworkMetadataLocker() {
+            if (m_ArtworkMetadata != nullptr) {
+                m_ArtworkMetadata->release();
+            }
+        }
+
+    public:
+        ArtworkMetadata *getMetadata() const { return m_ArtworkMetadata; }
```
Ribtoks: I would suggest name `getArtwork()` or `getArtworkMetadata()`.
Just `getMetadata()` is a bit confusing because of class `BasicMetadataModel` which is often referred as "metadata".

src/xpiks-qt/Models/sessionmanager.cpp
```diff
@@ -0,0 +1,149 @@
...
+        QJsonDocument &doc = m_Config.getConfig();
+        if (doc.isObject()) {
+            m_SessionJson = doc.object();
+        } else {
+            LOG_WARNING << "JSON document doesn't contain an object";
+        }
+    }
+
+    void SessionManager::saveToFile(std::vector<MetadataIO::ArtworkMetadataSnapshot *> &snapshot) {
+        LOG_DEBUG << "#";
```
Ribtoks: Might be useful to log the number of artworks to save

src/xpiks-qt/Models/sessionmanager.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+        m_Mutex.lock();
+        m_Config.saveToFile();
+        m_Mutex.unlock();
+    }
+
+    void SessionManager::restoreFromFile() {
+        LOG_DEBUG << "#";
+
+        const QJsonArray &filesArray = value(OPENED_FILES).toArray();
```
Ribtoks: Method `restoreFromFile()` actually restores it from memory. Quite confusing naming/behavior.

Why do you always keep it in memory, btw? It is already in the `ArtItemsModel` in the instantiation of vector of artworks. If it's not there, it should be just read from disk. I mean, shouldn't it be read from disk when you call `restoreFromFile`?

src/xpiks-tests/xpiks-tests-integration/restoresessiontest.cpp
```diff
@@ -0,0 +1,134 @@
...
+    if (!waiter.wait(20)) {
+        VERIFY(false, "Timeout exceeded for reading session metadata.");
+    }
+    VERIFY(!ioCoordinator->getHasErrors(), "Errors in IO Coordinator while reading");
+
+    MetadataIO::SessionSnapshot newArtworksSnapshot(artItemsModel->m_ArtworkList);
+    auto oldArtworksList = oldArtworksSnapshot.getSnapshot();
+    auto newArtworksList = newArtworksSnapshot.getSnapshot();
+
+    VERIFY(oldArtworksList.size() == newArtworksList.size(), "Old and new snapshots has different number of items");
```
Ribtoks: SnapshotS _have_, not _has_

PR #519, opened 2018-01-11
Add project for build libthmbnlr, and fix DESTDIR for hunspell.

velmafia:
Hi,
I'am new here, so I started from build, and I found some issue (?):
1. Default compilation is using lthmbnlr, but there is no project to build it, so I create simple project for that.
2. Also I fixed desination directory for hunspell release target.

aloha
   Kamil


ribtoks:
Thanks @velmafia

PR #528, opened 2018-01-19
Upgrade build_vendors_travis.sh to be usable also for local vendors preparation in unix base system.

velmafia:
I don't like run manually what can be run automatically, so I upgrade build_vendors_travis.sh for unix base system, here is my proposition (pre-review).

My questions:
1. Why travis-ci target don't use libthmbnlr and translations (_src/xpiks-qt/deps/translations/_)?
2. Where it should be default libs directory, `libs`, or `libs/debug` and `libs/release`.

aloha!

PR #530, opened 2018-01-31
Libs consistency, and translation copy

velmafia:
Hi,
according comments in #528, I unified libs location for xpiks.
Also I change condition for translation copy, as you propose.  

aloha!


ribtoks:
Thank you!

PR #557, opened 2018-05-11
quick fix for race condition in loger emergency flush. 

velmafia:
In my linux environment integration tests run is aborted due the race condition in logger in more that 50% case (based on 66e9654a7cfcd10db4b7afb76b954a3869f806bb commit)
`Logger::emergencyFlush()` don't use mutex guard, so `QStringList *logItems` in `Logger::flushStream(...)`  is corrupted when test prints warning log.


ribtoks:
@velmafia maybe the morning is too early, but would you be so kind to explain how the `logItems` inside `flushStream()` gets corrupted?

ribtoks:
@velmafia ok, you mean crashes the one which is before `abort()` in `myMessageHandler()` not in the real crash handler.. Now I see the problem. In case if it's our "graceful" crash, logging thread is still alive.

PR #582, opened 2018-09-30
adjust xpiks code to fixed includes in cpp-libface

velmafia:


ribtoks:
I will approve after CI turns green

PR #583, opened 2018-10-01
Cleanup includes: vendors, xpisk-qt/Artworks

velmafia:
* Change vendors include to full path
* Proposal of includes cleanup: `xpiks-qt/Artworks`


src/xpiks-qt/Artworks/artworkmetadata.h
```diff
@@ -11,35 +11,52 @@
...
-#include <Common/flags.h>
-#include <Common/types.h>
-#include <Common/delayedactionentity.h>
-#include <Services/SpellCheck/spellcheckinfo.h>
-#include <Services/SpellCheck/ispellcheckable.h>
-#include <UndoRedo/artworkmetadatabackup.h>
+#include <QString>
+#include <QStringList>
+#include <QWriteLocker>
+#include <Qt>
```
ribtoks: Why this include is needed?

src/xpiks-qt/Artworks/artworkmetadata.h
```diff
@@ -11,35 +11,52 @@
...
+#include <QtGlobal>
+
+#include "Artworks/basicmetadatamodel.h"
```
ribtoks: Why style of Xpiks includes does not start with `<`?

src/xpiks-qt/Artworks/artworkssnapshot.cpp
```diff
@@ -8,7 +8,15 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-#include "artworkssnapshot.h"
+#include "Artworks/artworkssnapshot.h"
```
ribtoks: Why should .cpp include .h with a special path rather than just without a path?

src/xpiks-qt/Artworks/videoartwork.h
```diff
@@ -11,13 +11,26 @@
...
 #include <QMutex>
+#include <QObject>
+#include <QReadLocker>
+#include <QReadWriteLock>
 #include <QSize>
-#include "../Common/flags.h"
-#include "../Common/types.h"
+#include <QString>
+#include <QWriteLocker>
+#include <QtGlobal>
```
ribtoks: Why do you need to include `QtGlobal` every time? To my knowledge it is included with every `<QSomething>`

src/xpiks-qt/Connectivity/ftphelpers.cpp
```diff
@@ -9,7 +9,7 @@
  */
 
 #include "ftphelpers.h"
-#include <uploadcontext.h>
+#include <vendors/libxpks/uploadcontext.h>
```
ribtoks: I'm not sure I follow the logic: in some places you put spaces to separate groups and in the others you don't.

src/xpiks-qt/MetadataIO/metadataiocoordinator.cpp
```diff
@@ -18,8 +18,8 @@
 #include "../Models/settingsmodel.h"
 #include "../Common/defines.h"
 #include "metadataioservice.h"
-#include <readingorchestrator.h>
-#include <writingorchestrator.h>
+#include <vendors/libxpks/readingorchestrator.h>
```
ribtoks: There're 2 implementation of `libxpks` and with this way of including you will always eliminate `src\libxpks_stub` implementation.

src/xpiks-qt/Artworks/artworkmetadata.h
velmafia: This include is for `Qt::TimerType`. As `iwyu` suggest (and [Qt doc](http://doc.qt.io/qt-5/qt.html)), `<Qt>` should be include for this type

src/xpiks-qt/Artworks/videoartwork.h
velmafia: This include was suggested by `iwyu`: `// for Q_UNUSED, qint64`.
This is general rule to not rely on "indirect includes", maybe in this case this isn't necessary.
But in my experience it's better to have some "correct, but not necessary includes", than rely on indirect includes, so I included it.

velmafia:
> now it's impossible to include src/libxpks_stub

You are right, I will think about it.

> why should first include in .cpp specify header by path rather than without?

As we talked by email, I use [Google C++ Style](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes). In my experience this is the most clear convention, but of course this is a matter of contract.

>AppVeyor build is red for some reason - need to investigate

In my branch `AppVeyor` was [green](https://ci.appveyor.com/project/velmafia/xpiks/build/1.5.38), this look like environment problem. Maybe rebuild will help?
> `ERROR: The process "exiftool.exe" not found.`

src/xpiks-qt/Artworks/artworkssnapshot.cpp
velmafia: Please see general comment.

src/xpiks-qt/Artworks/videoartwork.h
velmafia: Additional: http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-implicit

src/xpiks-qt/Artworks/artworkmetadata.h
velmafia: As I wrote in general comment, I use [Google C++ Style](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes).

In my opinion it's nice rule to show in includes list what headers is "our", and what file are include outside.

---------------
Additional:
Of course we can use both notation alternatively, but general practice is (as I know):
* `"..."` notation - for "local" project headers
* `<...>` notation - for system/external headers 

For example in [gcc](https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html):
>For the angle-bracket form #include <file>, the preprocessor’s default behavior is to look only in the standard system directories.

ribtoks:
@velmafia "process exiftool.exe not found" error is not related - it's an artifact of "kill exiftool.exe" in the .bat script. The real reason is failling `CsvExportTest`
```
01:36:12.225 info T#1200 main - --------------------------
01:36:12.225 info T#1200 main - Integration Tests Results: 45 succeeded, 1 failed
01:36:12.225 info T#1200 main - Tests return code 2
01:36:12.225 info T#1200 main - FAILED TESTS: "CsvExportTest"
01:36:12.225 info T#1200 main - --------------------------
```

Something is fishy there..

Regarding google c++ style, I agree to everything `#include`-wise except of current .cpp file's header. It helps a lot to find it visually if it is not included using path.

ribtoks:
@velmafia Also, with this problem with `libxpks_stub`, how travis-ci build turned green? Supposedly it shouldn't be able to find includes.. Overall - we can we move/rename `libxpks_stub`, that's not a problem. Just it should be nice and natural.

velmafia:
> Regarding failing `CsvExportTest`

I will check this.
> Regarding google c++ style, I agree to everything #include-wise except of current .cpp file's header. It helps a lot to find it visually if it is not included using path.

OK, I will change includes for current .cpp

> Also, with this problem with libxpks_stub, how travis-ci build turned green? Supposedly it shouldn't be able to find includes.

Headers from `libxpks_stub` is included only in `libxpks_stub` (from the same dir) and in `vendors/libxpks`. This is reason why CI is green.  

If the only "interface" to `libxpks_stub` is `vendors/libxpks`, so where we can move/rename `libxpks_stub`?  
* move `libxpks_stub` as "vendor library"?
* or build as "internal lib"?
* other?
What do you think? (I really don't like exceptions  in code;) )

src/xpiks-qt/Artworks/artworkmetadata.h
ribtoks: Ok, agree

src/xpiks-qt/Artworks/videoartwork.h
ribtoks: Ok, let it be there

src/xpiks-qt/Connectivity/ftphelpers.cpp
ribtoks: OK, that's understandable. Just looks inconsistent a little bit, but I didn't notice file was not from `Artworks`

src/xpiks-qt/MetadataIO/metadataiocoordinator.cpp
ribtoks: Sorry, I mislead you. Actually I put includes from `libxpks` to `/vendors/` long time ago, so this is not an issue.

ribtoks:
@velmafia Agree about exceptions. It's good that you have same opinion.

Actually includes from `libxpks` are already in vendors, so there's no need to move it there. Just build-time library is taken from local or remote source. I guess that there's no problem in this case, sorry for the confusion.

There's still a problem with red AppVeyor build. If this is still CSV export test, you can open integration tests project, comment out all other tests in `main.cpp` and debug it locally. The test is pretty simple, just remember to put `exiftool.exe` somewhere where integration tests expect to find it (current dir, PATH, Downloads dir, don't remember what else).

velmafia:
I change headers includes for cpp file. Also I try tu reproduce fail in `CsvExportTest` - on Linux because I don't have Windows system. Unfortunately in my local env everything is working, also in my [travis](https://travis-ci.org/velmafia/xpiks/builds/436884310) and [appveyor](https://ci.appveyor.com/project/velmafia/xpiks/build/1.5.41). 
But after commit [travis fail](https://travis-ci.org/ribtoks/xpiks/builds/436884326?utm_source=github_status&utm_medium=notification)... I can't reproduce it in my local system, and also in [my travis](https://travis-ci.org/velmafia/xpiks/builds/436884310) everything works... 
Do you meet before some race condition in this part of code?

ribtoks:
@velmafia if you take a look at Travis-CI, it was just UI tests that failed (because of timings), not Csv test. If this happens again you can just restart the job (apparently container that was running that build was extra slow). PR indicates that AppVeyor (Windows) build succeeded so it's OK. But why it failed before - that's a good question. Theoretically there could be a race condition in CSV exporting - if you have time you can take a look inside (first - we need to understand what condition exactly is making test red).

PR #586, opened 2018-10-07
Cleanup includes in xpisk-qt - next part

velmafia:
- Cleanup includes in src/xpiks-qt/Common
- Cleanup includes in xpiks-qt/Encryption
- Cleanup includes in xpiks-qt/Commands/Base
- Cleanup includes in xpiks-qt/Commands/Editing
- Cleanup includes in xpiks-qt/Commands/Files
- Cleanup includes in xpiks-qt/Commands/Services
- Cleanup includes in xpiks-qt/Commands/UI
- Cleanup includes in xpiks-qt/Commands
- Cleanup includes in xpiks-qt/Connectivity
- Cleanup includes in xpiks-qt/Filesystem 
- Cleanup includes in xpiks-qt/Helpers


velmafia:
IWYU have problem with `#ifdef`'s, I fix win and mac compilation today.

src/xpiks-qt/Commands/Editing/expandpresettemplate.h
```diff
@@ -53,7 +59,7 @@ namespace Commands {
         virtual void execute(Artworks::ArtworksSnapshot const &snapshot) override;
 
     private:
-        int m_CompletionID;
+        int m_CompletionID;  // this value is unused - it should be removed?
```
ribtoks: Not in this pull request

ribtoks: Please next time add a comment in PR description or in chat instead of in code.

src/xpiks-qt/Common/systemenvironment.cpp
```diff
@@ -9,14 +9,23 @@
...
+#include <QtDebug>
+#include <QtGlobal>
+
+#include "Common/defines.h"
+#include "Common/logging.h"
+#include "Helpers/constants.h"
+#include "Helpers/filehelpers.h"
+
+#if defined(INTEGRATION_TESTS) || defined(UI_TESTS)
 #include <QDateTime>
```
ribtoks: Is `<QDateTime>` needed only for tests? Can it be included with all Qt includes in this case?

src/xpiks-qt/Common/types.h
```diff
@@ -11,6 +11,7 @@
 #ifndef TYPES_H
 #define TYPES_H
 
+#include <algorithm>
```
ribtoks: It doesn't make any sense to include `<algorithm>` in `types.h`. The latter is used as "low-level" include with only a couple of things. If it will bring `<algorithm>` every time it will slow down compilation time significantly.

src/xpiks-qt/Connectivity/simplecurldownloader.cpp
```diff
@@ -9,13 +9,20 @@
...
 #include <QTemporaryFile>
-#include <cstdio>
+#include <QtDebug>
+#include <QtGlobal>
+
 #include <curl/curl.h>
-#include <Common/logging.h>
-#include <Models/Connectivity/proxysettings.h>
-#include "ftphelpers.h"
+#include <curl/curlver.h>
```
ribtoks: No need to include `curlver.h`. Curl documentation says all curl API is supposed to be used from `curl.h`

src/xpiks-qt/Connectivity/uploadwatcher.h
```diff
@@ -12,8 +12,17 @@
...
 #include <QAbstractListModel>
-#include <QVector>
+#include <QHash>
+#include <QObject>
 #include <QPair>
+#include <QString>
+#include <QStringList>
+#include <QVariant>
+#include <QVector>
+#include <Qt>
```
ribtoks: Doesn't make sense to include `<Qt>` in the end. With probability almost 100% everything which is needed from there has been transitively included from other files. Either put it in front, or don't use it at all. There's no chance in the universe Qt guys will not include it from `<QObject>` or any other always-included file.

src/xpiks-qt/Encryption/aes-qt.cpp
```diff
@@ -9,10 +9,15 @@
...
-#include <QDataStream>
+
+#include <cstdint>
+
+#include <QByteArray>
+#include <QChar>
 #include <QCryptographicHash>
-#include <cstdio>
+#include <QString>
+#include <QtGlobal>
```
ribtoks: Same comment regarding `<QtGlobal>` as for `<Qt>` above.

src/xpiks-qt/Encryption/secretsmanager.cpp
```diff
@@ -9,10 +9,17 @@
...
-#include "aes-qt.h"
+
+#include <QChar>
+#include <QCharRef>
 #include <QCryptographicHash>
+#include <QMutexLocker>
 #include <QTime>
-#include "../Common/logging.h"
+#include <QtDebug>
+#include <QtGlobal>
```
ribtoks: `<Qt>` comment here..

src/xpiks-qt/Commands/Editing/expandpresettemplate.h
velmafia: OK.

velmafia:
Hi, 

As we talk in #583, I use [Google C++ Style](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes), and this is reason why `<Qt>`, `<QtGlobal>` and another general include aren't before specific includes -  [Google C++ Style](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) recommended alphabetically order.

> Doesn't make sense to include `<Qt>` in the end. With probability almost 100% everything which is needed from there has been transitively included from other files. Either put it in front, or don't use it at all. There's no chance in the universe Qt guys will not include it from `<QObject>` or any other always-included file.

Yes, you are right, but you yourself said `probability almost 100%` - not simply say `everything will be included`. What will be included in `<QObject>` is implementation detail of `QObject`, I think it's better not to rely on `Qt` implementation details. Of course, in this case this will be always true, but what about another includes? I really don't like exceptions in general rules, so I'd rather leave it as it is. Without exceptions for general/specific includes. 
If include was included before by another includes - it will nothing changed, include guards will simply cut its body, and we don't lose independence from previous includes. Of course we can slightly slow down compilation time, but includes guard are efficient, and simple rule without exception ensure easier adherence to this rules. And finally cleaner includes reduce compilation time (see below), and  reduce unexpected behaviour. 
 
All above brings to [this rule](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-implicit). And I strongly agree with that (this is my private historical reason - experience in work...).

---------------
About compilation time:
After my changes compilation time is reduced from  `(~11-13min)` to `(~10:30 min )`, you can compare compilation time of my branch to master on `travis-ci`. I have similar experience from another really big project - includes cleanup decrease compilation time about 20-30%. In this project I also added a lot of _correct but unnecessary_ includes, but include guard is match more faster that parsing unnecessary includes. 

For this reason I prepare simple script for automatic running `iwyu` for xpiks project. If you want I can connect it to `travis-ci` after cleanup, but for that all files need have all used includes, and nothing more. This is main reason why I don't want any exception. Also alphabetically order nothing changes in finally binary, it's easy/fast to maintenance (I think all IDE, editor have option for sort lines), and alphabetically order is easy to optical search something.

src/xpiks-qt/Encryption/aes-qt.cpp
velmafia: Please see general comment.

src/xpiks-qt/Encryption/secretsmanager.cpp
velmafia: Please see general comment.

ribtoks:
Hello

I clearly remember that your incentive was to introduce Google C++ Style (at least for includes) and all what you did comes from that incentive, I guess no need to remind me. I must warn you that I do not want to bring thing just because they are from Google or because 1000 other projects do that. I want to bring things/ideas in only of they bring value. For example, in my understanding including small broad includes (like `<Qt>`) before others brings value (in terms of speed and iwyu) but including it in alphabetical order does not. I don't care if it is recommended by Google or not. I try to use facts and brains before using a feature.

If you can give me at least 1 good (except "recommended by google") reason to use includes in alphabetical order but not from small&common to specific, I'm all ears. I see good reasons in global order (c++ std, qt std, custom libs, xpiks codebase), but not in alphabetical order yet (unless you can give some reason to me).

Next, regarding compilation time. Please do not make premature conclusions about impact of includes reordering. It's not that I want to diminish boost of this PR, just in my experience xpiks-qt binary build on Travis used as much as 8 to 13 minutes to compile. This range exists most probably due to the load of current given server that was running that compilation. If you want to make a perftest and declare compilation boost, it should be tested in a same controllable environment (like your computer, not like common buildserver).

Re: Qt includes again: I've been using Qt from 5.2 to 5.11. You're claiming that the fact that `<QObject>` includes `<Qt>` is an implementation detail and I agree. Even if chances that this will change are almost 0, yes, it is an implementation detail. But what you suggest to include `<Qt>` because it contains some enum is also an implementation detail. Qt Documentation is not set in stone. They might move the enum or define to the different include file and change the documentation. Then this change will be equivalent to not including `<Qt>` from `<QObject>` and equal to the same result: necessity of doing changes in projects that are using Qt.

velmafia:
Hi,
about alphabetically order, I don't think that this is value because it's from google, sometimes I have problem to clearly write what I think.

So maybe I write my point if view:
As I know C++ (if I'm not right - please correct me), order of includes have absolute no matter (of course if included files have correct include guards, and if are included in global namespace). All included directive will be replaced by its file body only once. If I move `<QtGlobal>` at the beginning of includes list, it will be included first, and will be ignored later in another headers. If it will be included last, it will be included by another header, and at the end of includes list it will be ignored. (AFAIK) regardless of the order of includes, include guard will be check the same number of times, only "include deep" and order will be changed (in c++ order of symbols don't have matter, because they must be unique in translation unit). Output file will be [exactly the same](https://pastebin.com/jPSukN2u).
Also I make some benchmark with `-ftime-report` for `xpiksapp.cpp`, and output was very similar (differences on the noise level).

Alphabetically order have one advantage - it is simple and easy to maintain. Also it is indisputable (nobody denies the alphabet order), and general-specyfic order can be understood differently by different people. So it's bigger chance to make code/includes consistent. 
IMO pros of alphabetically order:
* simply rule;
* easy maintain;
* consistent independent of developers;
* easy merge (usually by automerge);
* easy "eye searching";

But this in only my preference, and you are boss here :)

------
About compilation time: 
Travis comparison was some shortcuts from my side - sorry for that. I make some test in my computer, and travis time was similar. So I decide to not prepare better comparison - I'm doing it right now ;)

All test was made with `-j4` flags.

|itereation|master| includes-xpiks-qt_part2|include-cleanup (my local branch) |
|----------|------|------------------------|----------------------------------|
|1| real    4m15,871s <br> user    15m36,802s <br> sys     1m24,652s| real    4m1,178s <br> user    14m36,665s <br> sys     1m17,934s | real    3m55,132s <br> user    14m11,281s <br> sys     1m16,558s |
|2| real    4m17,071s <br> user    15m37,951s <br> sys     1m23,443s| real    4m4,436s <br> user    14m42,719s <br> sys     1m20,294s | real    3m54,567s <br> user    14m9,597s <br> sys     1m16,699s |
|3| real    4m16,868s <br> user    15m36,556s <br> sys     1m23,839s| real    4m2,785s <br> user    14m40,899s <br> sys     1m19,244s | real    3m55,101s <br> user    14m10,902s <br> sys     1m16,706s |
|4| real    4m14,472s <br> user    15m33,019s <br> sys     1m22,446s| real    4m3,717s <br> user    14m40,390s <br> sys     1m19,975s | real    3m56,382s <br> user    14m15,332s <br> sys     1m17,935s |
|5| real    4m17,547s <br> user    15m39,946s <br> sys     1m23,113s| real    4m3,021s <br> user    14m41,130s <br> sys     1m19,938s | real    3m54,552s <br> user    14m10,166s <br> sys     1m14,631s |

So compilation of current PR version is ~6-7% faster, and my local branch is ~10% faster.

I don't commit `include-cleanup` branch yet, because my cleanup in `xpiks` project [break](https://travis-ci.org/velmafia/xpiks/jobs/445423604) `xpiks-helloworld-plugin` project :/ This is another reason why I want cleanup this includes.

In this branch I clean standalone headers, because iwyu was ignore it before (`--check_also` swith not works as doc say...).

-----
About "Qt includes":

My understanding of C++: this enums is part of "public interface" of qt framework - so it's not implementation detail, also it's usage are documented. Includes are not documented, and depends of internal `Qt` implementation, so they can be changed without inform user about this changes.

aloha!

src/xpiks-qt/Connectivity/simplecurldownloader.cpp
velmafia: OK

ribtoks:
Hi

Thanks for digging it up and producing cool results (in terms of compilation speed).

Re: includes. First of all, +1 for point with merging (easy to maintain and eye searching are questionable I would say, for me they are same right now). Secondly, regarding speed/order/etc. - my point with most common includes was they since they are most common - all subsequent includes most probably include them as well. But I can agrue now with myself that probably first transitive include of most common one with still include it and theoretically should be no difference in speed in this regard.

Re: Qt. Pls read again my point in a post before. I said that documentation is not set in stone (actually I knew you would write what you wrote and tried to reply to this next post). Even if in this version the doc says enum is in one header it can also change. If you include `<QObject>` and it transitively includes enum - you kind of get the same result (with the excuse that it's bad way to achieve it). Next version both facts can change. But it's more philosophical discussion, probably will need to stop it.

P.S. Found a [post by a dude](https://blog.knatten.org/2010/07/01/the-order-of-include-directives-matter/) who recommends doing reverse as Google does in terms of global ordering (you, 3rdparty libs, std libs) because if you're not using iwyu and you made unfortunate mistake of using some implicit include his approach will show it. I see a very valid point, but it's kind of late to change stuff now.

PR #589, opened 2018-10-25
Cleanup standalone headers in xpiks-qt project

velmafia:
* Additional cleanup of stand-alone headers (default `iwyu` don't check this headers);
* Some additional clean;


ribtoks:
Why these headers are "standalone"?

velmafia:
"standalone" I mean every headers that does't have corresponding cpp file (for example interface headers).

From iwyu doc:
> Include-what-you-use only analyzes .cc (or .cpp) files built by make, along with their corresponding .h files. If your project has a .h file with no corresponding .cc file, IWYU will ignore it unless you use the --check_also switch to add it for analysis together with a .cc file.

`--check_also` not working as doc say ([not only for me](https://github.com/include-what-you-use/include-what-you-use/issues/268)), and in previous commits I didn't clean this file.

PR #599, opened 2018-11-19
Disable "Title text input" in "Preset menu" when preset list is empty.

velmafia:
Disable "Title text input" in "Preset menu" when preset list is empty - crash fix.


src/xpiks-qt/Dialogs/PresetsEditDialog.qml
```diff
@@ -383,13 +384,15 @@ Item {
                             }
 
                             onEditingFinished: {
-                                if (text.length == 0) {
-                                    if (presetNamesListView.currentItem) {
-                                        presetNamesListView.currentItem.myData.editname = qsTr("Untitled")
+                                if(presetNamesListView.currentItem) {
+                                    if (text.length == 0) {
+                                        if (presetNamesListView.currentItem) {
```
ribtoks: Why there're now two `if (presetNamesListView.currentItem)` conditions?

velmafia: Of course only one `if` is required. 
Next time I will not open PR at 4:00 o'clock...

Issue #693, opened 2019-05-25
Suggest keywords don't search editorial images in shutterstock


velmafia:
Shutterstock API doesn't return any of editorial images for `/v2/images/search` endpoint.

`license` parameter should be added to search query: https://api-reference.shutterstock.com/?shell#images

PR #694, opened 2019-05-25
Add editorial images to "Suggest keywords" for shutterstock.

velmafia:
Proposal for #693


ribtoks:
Few questions:
- did you try if it works?
- why do you think somebody wants to see editorial images (maybe, except you)?

velmafia:
> * did you try if it works?

Yes, I try it and it works.

> what about other microstocks where Xpiks supports search (Fotolia/Adobe, iStock)?

* iStock/Getty -  Image currently return also editorial image;
* Fotolia/Adobe-  as I know, Fotolia never accepted editorial content, so probably API don't support it (currently I can't check official documentation, because of forwarding Fotolia to Adobe Stock [link](https://pl.fotolia.com/Services/API/Rest/Documentation) is broken). Adobe Stock don't accept editorial artworks from individual contributor, only from they partnership like Reuters. 

> * why do you think somebody wants to see editorial images (maybe, except you)? For example, I don't have any data (user requests on forums/support/emails) that suggests anybody needs this option.

Personally I don't take typically editorial photos, but some photo are accepted only as editorial (cityscape for example - because of shops facade, etc). 
Also, probably `xpiks` user didn't know that `suggest keywords` tool doesn't return editorial content for suggestion, because of popular topic have a lot of royalty-free images to fill suggestions thumbnail. 
I found this `xpiks` behaviour when I try tagging some niche topic - I found some photos on shuttestock, but xpiks return nothing. So I try shutterstock [web keyword suggestion tool](https://submit.shutterstock.com/portfolio/suggestions.mhtml?language=en) - they return this niche images for suggestions.

> * did you evaluate any alternative options to achieve same result? (e.g. if user adds a search term "editorial" together with real search terms, Xpiks can set "editorial" license type)

* I think returning both type of image is better option (as in shutterstock web tool), because similar images you can find in both category.  
* Separate searching for editorial/non-editorial images may cause the need to merge result manually.
* I want by consistent with currently `iStock` behaviour.

ribtoks:
Overall I understand the intent and your change is rather trivial, but I still need to think about it for some time. The thing is that everybody mainly use Shutterstock keywords suggestion (not much of fotolia and getty) so we need to be careful with changing default behavior for majority of the users.

velmafia:
I understand your doubts, but I want to present my POV:

The main difference between `editorial` / `non-editorial` images is them licence, not image content. You can find very similar  content in both category (for example try search `lviv` or `wroclaw` in shutterstock - half of result is editorial). Keyword for all this images are almost the same, no matter of licence type. Also rest of typical editorial content (like events) have very similar keywords in `editorial` - `non-editorial` images - only difference is if author grab all necessary `property release` or not.

Finally, user chooses what images are similar to his artworks, so I think will be better show him set of more popular images in both category (for example, for `lviv` keywords, 8 of 20 first images are editorial). I suppose that shutterstock itself has a similar opinion, and this is reason why they return suggestions from both category.

ribtoks:
Ok, merged. Thanks for PR!

PR #712, opened 2019-06-08
Don't use local config for switches, if it is older than 14 days.

### LINKED ISSUE #672: Expire Switcher config after few days
Save to switcher state something like "last updated datetime". On loading local switcher config if the difference between "now" and "last" is more than 7 days, fallback to switcher defaults.

velmafia:
close #672


src/xpiks-qt/Models/switchermodel.cpp
```diff
@@ -115,6 +122,30 @@ namespace Models {
...
+    void SwitcherModel::checkLocalConfigExpired()
+    {
+        bool expired = false;
+        const QString lastUpdate = m_State.getString(LAST_UPDATED_DATETIME);
+        const QDateTime lastUpdateDateTime = QDateTime::fromString(lastUpdate, Qt::ISODate);
+
+        do {
+            if (!lastUpdateDateTime.isValid()) {
+                LOG_WARNING << "Cannot parse last update daytime:" << lastUpdate;
+                expired = true;
```
ribtoks: I'm not sure I can agree with this logic. If we don't have date in the config, it won't parse as well as valid. Let's not do any decisions if we don't have in the "state". In worst case this will cost us 1 more day. If we never had any remote config, we will use defaults anyway.

src/xpiks-qt/Models/switchermodel.cpp
```diff
@@ -115,6 +122,30 @@ namespace Models {
...
+                break;
+            }
+
+            const QDateTime dtNow = QDateTime::currentDateTime();
```
ribtoks: Please use `currentDateTimeUtc()` and always use UTC in any DateTime calculations.

src/xpiks-qt/Models/switchermodel.cpp
```diff
@@ -115,6 +122,30 @@ namespace Models {
...
+            qint64 daysPassed = lastUpdateDateTime.daysTo(dtNow);
+            if(daysPassed > SWITCHER_EXPIRATION_DAYS) {
```
ribtoks: [cosmetic]
I would just replace this `while` with 
```
const bool expired = !lastUpdate.empty() && 
                      lastUpdateDateTime.isValid() &&
                      lastUpdateDateTime.daysTo(dtNow) > SWITCHER_EXPIRATION_DAYS;
```

src/xpiks-qt/Models/switchermodel.cpp
```diff
@@ -136,6 +167,15 @@ namespace Models {
         emit switchesUpdated();
     }
 
+    void SwitcherModel::switcherArrived() {
+        LOG_DEBUG << "#";
+
+        const QDateTime dtNow = QDateTime::currentDateTime();
```
ribtoks: Please use `currentDateTimeUTC()`

src/xpiks-qt/Models/switchermodel.cpp
```diff
@@ -37,6 +39,9 @@ namespace Models {
         QObject::connect(&m_Config, &Connectivity::SwitcherConfig::switchesUpdated,
                          this, &SwitcherModel::configUpdated);
 
+        QObject::connect(&m_Config, &Connectivity::SwitcherConfig::switcherArrived,
+                         this, &SwitcherModel::switcherArrived);
```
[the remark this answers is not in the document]
velmafia: done

src/xpiks-qt/Models/switchermodel.cpp
```diff
@@ -115,6 +122,30 @@ namespace Models {
         }
     }
 
+    void SwitcherModel::checkLocalConfigExpired()
+    {
```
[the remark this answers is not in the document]
velmafia: done

src/xpiks-qt/Models/switchermodel.cpp
velmafia: done

src/xpiks-qt/Models/switchermodel.cpp
velmafia: done.

Change this also for code that already exist in this file?
- SwitcherModel::ensureSessionTokenValid()
- SwitcherModel::getDonateCampaign1LinkClicked()

src/xpiks-qt/Models/switchermodel.cpp
velmafia: done

src/xpiks-qt/Models/switchermodel.cpp
velmafia: done

Issue #691, opened 2019-05-22
Create switch to disable legacy backups cleanup


ribtoks:
Start legacy backups cleanup MaintenanceTask only if the appropriate switch is ON.
To create a switch see an example of anything existing in `SwitcherModel` (basically you need to copy-paste few lines of code in `SwitcherConfig` for declaration and parsing and pass-through it as a method in `SwitcherModel`)

velmafia:
I'm working on it.

editing `switches.json` in branch `gh-pages` is also in scope in this task?

What should be default value of this switch: `legacy backups cleanup` enabled or disabled?

ribtoks:
Default value should be `disabled` (it will be disabled for all newer Xpiks releases like 1.6+ since "old backups" is a thing from past).

If it's not hard, yes, you can update `switches-153.json` as well.

PR #715, opened 2019-06-10
Add switcher for "legacy backups cleanup".

velmafia:
related #691


ribtoks:
@velmafia default value should be false in C++, but it should be turned ON in json for now. Thanks

PR #716, opened 2019-06-10
Add switch for "legacy backups cleanup".

### LINKED ISSUE #691: Create switch to disable legacy backups cleanup

velmafia:
close #691


ribtoks:
@velmafia that's cool, but please also make use of this key :)
It's used in `CleanupLegacyBackupsCommand` - add `SwitcherModel` reference as a member and run cleanup only if it is allowed.

velmafia:
done :)

Issue #475, opened 2017-09-24
Preset smart groups


ribtoks:
Recently used
Recently added

Create these groups using `RecentItemsModel` and deriving Presets Manager from `StatefulEntity` and saving recent groups to state.

Related #474

PR #721, opened 2019-06-19
Issues/475 smart groups

velmafia:
Proposal for #475 

Suggestion from https://github.com/velmafia/xpiks/commit/c8803b3bd6807609cd1eed6dce96c9fc100beaf included.


src/xpiks-qt/Common/lrucache.h
```diff
@@ -48,6 +48,15 @@ namespace Common {
             }
         }
 
+        bool contains(const TKey &key) const {
```
ribtoks: That's pretty much a duplication of code of `tryGet()`. So far your `contains()` is only used in 1 place. Probably it is not worth of a separate copy-paste method.

src/xpiks-qt/KeywordsPresets/presetgroupsmodel.cpp
```diff
@@ -241,6 +246,32 @@ namespace KeywordsPresets {
         return result;
     }
 
+    QObject *PresetGroupsModel::getRecentlyUsedModel()
+    {
```
ribtoks: Please put brace on previous line to be consistent

src/xpiks-qt/KeywordsPresets/presetgroupsmodel.cpp
```diff
@@ -281,6 +312,29 @@ namespace KeywordsPresets {
         return changed;
     }
 
+    void PresetGroupsModel::refreshRecentlyGroup()
+    {
```
ribtoks: Put brace on previous line please

src/xpiks-qt/KeywordsPresets/presetgroupsmodel.cpp
```diff
@@ -281,6 +312,29 @@ namespace KeywordsPresets {
...
+        LOG_DEBUG << "#";
+        if (m_FilteredRecentlyUsed) {
+            m_FilteredRecentlyUsed->refresh();
+        }
+        if (m_FilteredRecentlyAdded) {
+            m_FilteredRecentlyAdded->refresh();
+        }
+    }
+
+    const std::map<int, int> PresetGroupsModel::getRecentlyAddedGroup() const {
```
ribtoks: There's no need for return parameter to be `const std::map`. Just `std::map` is enough

src/xpiks-qt/KeywordsPresets/presetkeywordsmodel.cpp
```diff
@@ -1008,4 +1020,18 @@ namespace KeywordsPresets {
 
         return result;
     }
+
+    RecentPresetsGroupModel::RecentPresetsGroupModel(const Common::LRUCache<int, int> &recentlyPresets, PresetKeywordsModel &presetsModel):
+        FilteredPresetsModelBase(presetsModel),
+        m_recentlyPresets(recentlyPresets) {
+        LOG_DEBUG << "Instantiated for recently group";
```
ribtoks: Not sure this is a useful log line

src/xpiks-qt/KeywordsPresets/presetgroupsmodel.cpp
```diff
@@ -281,6 +312,29 @@ namespace KeywordsPresets {
...
+        LOG_DEBUG << "#";
+
+        std::map<int, int> presetIdToPosition;
+        int pos = 0;
+        for (const auto &item : m_RecentlyAdded.data()) {
+            presetIdToPosition[item.first] = pos++;
```
ribtoks: I'd say that having `QDateTime` (when the preset was added) as I suggested in the preset is more generic and simpler IMHO. It allows you to deduct different things (recently added in particularly this case). The order you're saving is a little bit weird. So far I cannot make up some corner case, but I will think about it.

src/xpiks-qt/KeywordsPresets/presetgroupsmodel.h
```diff
@@ -72,19 +75,26 @@ namespace KeywordsPresets {
...
 
     public:
         const std::vector<GroupModel> &getGroups() const { return m_GroupsList; }
+        const std::map<int, int> getRecentlyAddedGroup() const;
+
         QStringList retrieveGroupNames();
         void initGroups(const std::vector<PresetGroupData> &groupsData);
         int getGroupID(size_t index) { Q_ASSERT(index < m_GroupsList.size()); return m_GroupsList[index].m_GroupID; }
+        Q_INVOKABLE void addRecentlyUsed(int presetID);
+        void addRecentlyAdded(int presetID);
```
ribtoks: This PR despite some super cosmetic issues has 1 bigger issue: where are the tests? :)

ribtoks: I prefer having more tests than less tests. Remember, if you think more tests don't fit into `easy` category, feel free to create new "trivial" issue and have 100% coverage for your change in tests there.

src/xpiks-qt/Common/lrucache.h
velmafia: I tried `tryGet()`, but this method also call `m_Data.splice(...)` what changes internal state. I use this method in `filterAcceptsRow(...)`and in this causes order of recently used/added is broken by UI internal mechanism. This is reason why I add this method.

src/xpiks-qt/KeywordsPresets/presetgroupsmodel.cpp
velmafia: done

src/xpiks-qt/KeywordsPresets/presetgroupsmodel.cpp
velmafia: done

src/xpiks-qt/KeywordsPresets/presetgroupsmodel.cpp
velmafia: done

src/xpiks-qt/KeywordsPresets/presetkeywordsmodel.cpp
velmafia: ok, changed to `LOG_DEBUG << "#";`

src/xpiks-qt/Common/lrucache.h
ribtoks: I see. Yes, makes sense then. But now you should also add tests for this method since everything in `common::` should be tested. There're existing tests for `LRUCache` in core tests

src/xpiks-qt/KeywordsPresets/presetgroupsmodel.cpp
velmafia: Adding `QDateTime` to `GroupModel::PresetData` I need add it also to `KeywordsPresets::PresetModel` (or make some extra weird mapping), because of `PresetKeywordsModelConfig::loadFromModel(...)` override all preset data. This increase data structure in `PresetKeywordsModel` which are not needed.

Also when I try this approach, there was more "weird changes` in `PresetKeywordsModel`, and this solution was simply simpler. But yes - I'm not proud of this part of code.

I have new idea how make this part based on `QDateTime`as you propose. I will try implement this and I add it as new iteration.

src/xpiks-qt/KeywordsPresets/presetgroupsmodel.h
velmafia: I also prefer more than less tests. But to be honest, this task was really painful for me, when I try some solution, it turned out that `xpisk` doesn't work as I expected, also this was my first meeting with `qml models` (and this code isn't really "Hello world" example :) ).

So I open this PR, because I wanted to know your opinion, if this is good approach or not, after that when you accept this direction, then I will write tests. 

---
OK, I create new "trivial" task for tests. and I will commit them to this PR, together with post-review changes.

src/xpiks-qt/Common/lrucache.h
velmafia: OK, I will do this.

src/xpiks-qt/KeywordsPresets/presetgroupsmodel.cpp
ribtoks: Adding new data member to `PresetKeywordsModel` or `PresetModel` is OK. I wanted to do it anyway for accounting but didn't have a valid reason. Now you have.

src/xpiks-qt/KeywordsPresets/presetgroupsmodel.h
ribtoks: Yeah, this is very far from "Hello world". Unfortunately all "Hello world" problems in Xpiks are more or less solved.

src/xpiks-qt/KeywordsPresets/presetgroupsmodel.cpp
ribtoks: So to be exact, if you will go with something like `QDateTime m_DateAdded`, you should add it to `KeywordsPresets::PresetModel` struct and also to `KeywordsPresets::PresetData` struct used for serialization. That is totally OK.

src/xpiks-qt/KeywordsPresets/presetgroupsmodel.h
ribtoks: Btw, you should ask more questions about "Xpiks doesn't work as you expected" - I can help you to either understand or fix things. Dealing with it alone all of the time - and you will only collect "pain", unless that is what you want.

velmafia: When I worked on this task, sometimes I was feeling that I understand what happens, but in fact - not really. 
Sometimes when I'm really noob in some topic the biggest question is what exactly should I ask :)
So probably when I start to "feel" `xpiks`, then I will have more questions.

src/xpiks-qt/KeywordsPresets/presetkeywordsmodel.cpp
```diff
@@ -412,9 +415,10 @@ namespace KeywordsPresets {
 
             beginInsertRows(QModelIndex(), lastIndex, lastIndex);
             {
-                m_PresetsList.emplace_back(std::make_shared<PresetModel>(nextID, name, keywords, DEFAULT_GROUP_ID));
+                m_PresetsList.emplace_back(std::make_shared<PresetModel>(nextID, name, keywords, DEFAULT_GROUP_ID, QDateTime::currentDateTimeUtc()));
             }
             endInsertRows();
+            m_GroupsModel.addRecentlyAdded(nextID);
```
ribtoks: I'm not sure this is a correct approach. There're plugins for Xpiks that register presets. This way when plugin loads it will always put itself in the "recently added". We should only add preset to recently added in first `if` branch, but not in `else` branch.

Also need to think how to handle the Date in this case. Maybe safer to set it back 10 years? I don't have a good answer at this stage. Probably you should leave a TODO in comments.

src/xpiks-qt/StackViews/CombinedEditView.qml
```diff
@@ -159,16 +159,58 @@ Rectangle {
...
+
             Instantiator {
                 model: presetsGroups
                 onObjectAdded: subMenu.insertItem( index, object )
                 onObjectRemoved: subMenu.removeItem( object )
                 delegate: Menu {
                     id: groupMenu
                     property int delegateIndex: index
-                    property var groupModel: presetsGroups.getGroupModel(index)
+                    property var groupModel: presetsGroups.getGroupModel( index )
```
ribtoks: Uhmm, why did you add these spaces? Code 5 lines above is an unfortunate artifact of copy-paste from Qt Wiki. Code convention here is like in C++, no need for anything exotic.

src/xpiks-qt/KeywordsPresets/presetkeywordsmodel.cpp
```diff
@@ -392,9 +394,10 @@ namespace KeywordsPresets {
 
             beginInsertRows(QModelIndex(), lastIndex, lastIndex);
             {
-                m_PresetsList.emplace_back(std::make_shared<PresetModel>(nextID, name, keywords, -1));
+                m_PresetsList.emplace_back(std::make_shared<PresetModel>(nextID, name, keywords, DEFAULT_GROUP_ID, QDateTime::currentDateTimeUtc()));
```
ribtoks: This already doesn't fit into the screen of GitHub review. Can you please add a newline before `std::make_shared`?

src/xpiks-qt/KeywordsPresets/presetkeywordsmodel.cpp
```diff
@@ -737,7 +743,29 @@ namespace KeywordsPresets {
...
+        LOG_DEBUG << "#";
+
+        m_GroupsModel.initGroups(m_PresetsConfig.getPresetGroupsData());
+
+        std::vector<std::pair<QDateTime, int>> recentlyAdded;
+        {
+            QWriteLocker locker(&m_PresetsLock);
+            Q_UNUSED(locker);
+            for (const auto &item : m_PresetsList) {
+                if(item->m_AddedDate.isValid()) {
```
ribtoks: Please put a space after `if`

src/xpiks-qt/KeywordsPresets/presetkeywordsmodel.cpp
```diff
@@ -737,7 +743,29 @@ namespace KeywordsPresets {
...
+                    recentlyAdded.push_back({item->m_AddedDate, item->m_ID});
+                }
+            }
+        }
+
+        std::sort(recentlyAdded.begin(), recentlyAdded.end(), std::greater<>());
+        for (int i = 0 ; i < RECENTLY_GROUP_SIZE && i < recentlyAdded.size() ; ++i) {
+            m_GroupsModel.addRecentlyAdded(recentlyAdded[i].second);
+        }
```
ribtoks: I would add a log line describing how many "recently added" you have available right now

src/xpiks-qt/StackViews/ArtworkEditView.qml
```diff
@@ -210,7 +256,9 @@ Rectangle {
                 delegate: MenuItem {
                     text: name
                     onTriggered: {
-                        artworkProxy.addPreset(subMenu.defaultGroupModel.getOriginalID(index));
+                        var presetID = subMenu.defaultGroupModel.getOriginalID( index )
+                        artworkProxy.addPreset( presetID )
+                        presetsGroups.addRecentlyUsed( presetID )
```
ribtoks: This PR is 99% perfect. However, this `Menu` will be impossible to test in UI tests. Also there will be a bug if you use preset autocompletion or expansion ([as here](https://xpiksapp.com/blog/2017/xpiks-14-detailed-features#presets)) - this method will not be called.

I propose to move method `addRecentlyUsed()` to all places where it is actually used.
* `ArtworkProxyBase::doAppendPreset`
* `ArtworkProxyBase::doExpandPreset`
* `ExpandPresetTemplate::expandPreset`

src/xpiks-qt/StackViews/CombinedEditView.qml
velmafia: Yes, I wanted to be consistent with rest of code. I will remove this spaces.

src/xpiks-qt/KeywordsPresets/presetkeywordsmodel.cpp
velmafia: What do you mean:

> We should only add preset to recently added in first if branch, but not in else branch.

`addRecentlyAdded(...)` is called here in `if` not in `else` branch.

Also I don't understand why we need set date back 10 years - do you mean with "plugin case"?

src/xpiks-qt/KeywordsPresets/presetkeywordsmodel.cpp
velmafia: done

src/xpiks-qt/KeywordsPresets/presetkeywordsmodel.cpp
velmafia: done

src/xpiks-qt/StackViews/ArtworkEditView.qml
velmafia: ok

src/xpiks-qt/KeywordsPresets/presetkeywordsmodel.cpp
velmafia: ok

src/xpiks-qt/KeywordsPresets/presetkeywordsmodel.cpp
ribtoks: Hm. It looks like it was some ~GitHub UI artifact~ misunderstanding from my side. I expanded all "diff" collapsed branches - looks OK, sorry.

src/xpiks-qt/KeywordsPresets/presetkeywordsmodel.cpp
```diff
@@ -737,7 +751,34 @@ namespace KeywordsPresets {
...
+    }
+
+    void PresetKeywordsModel::initGroups() {
+        LOG_DEBUG << "#";
+
+        m_GroupsModel.initGroups(m_PresetsConfig.getPresetGroupsData());
+
+        std::vector<std::pair<QDateTime, int>> recentlyAdded;
+        {
+            QWriteLocker locker(&m_PresetsLock);
```
ribtoks: You only need `QReadLocker`. Write lock of ReadWriteLock is needed only when you modify `m_PresetList` count

src/xpiks-tests/xpiks-tests-core/common_tests.cpp
```diff
@@ -36,6 +36,23 @@ void CommonTests::lruCacheGetMissingTest() {
...
 
+void CommonTests::lruCacheContainsKeyTest() {
+    Common::LRUCache<QString, int> cache(3);
+
+    int times = 15;
+    while (times--) {
+        cache.put(QString("test %1").arg(times), times);
+    }
+
+    QVERIFY(cache.contains("test 0"));
```
ribtoks: Your main point of `contains()` vs `tryGet()` was that `contains()` should not evict data from cache. I'd add 1 more test to verify that - pretty much copy-paste `CommonTests::lruCacheFrequentGetTest` and modify 
```

QCOMPARE(cache.get("test 14", -1), 14);

QCOMPARE(cache.get("test 0", -1), 0);
QCOMPARE(cache.get("test 1", -1), 1);
```

So that it does `contains()`, but not `get()`. Then the following three should not be `-1`:
```
QCOMPARE(cache.get("test 2", -1), -1);
QCOMPARE(cache.get("test 3", -1), -1);
QCOMPARE(cache.get("test 4", -1), -1);
```

velmafia: OK, I understand.

But in your example only `test 2` will not be `-1` :)

src/xpiks-qt/KeywordsPresets/presetkeywordsmodel.cpp
velmafia: done

src/xpiks-tests/xpiks-tests-core/common_tests.cpp
ribtoks: @velmafia no, if speaking precisely, even `test 2` will be `-1`, because it will be evicted by `test 14` in the `while` loop. But my point was about different. You have to adjust second keys of course.

ribtoks: I meant something like this:

```
while (times--) {
    cache.put(QString("test %1").arg(times), times);
    cache.contains("test 14");
    cache.get("test 13", -1);
}

int v;
QVERIFY(!cache.tryGet("test 14", v));
QVERIFY(cache.tryGet("test 13", v));
```

velmafia: OK, I will change this.

PR #730, opened 2019-06-28
Add option to open artwork with default application.

### LINKED ISSUE #620: Add option to Open artwork with default application
Add option "Open" next to "Show in folder" in the artwork's context menu that will do `open` on macOS, start process with filename on Windows and do smth like `xdg-open` on Linux

velmafia:
close #620


src/xpiks-qt/Helpers/helpersqmlwrapper.cpp
```diff
@@ -155,6 +159,23 @@ namespace Helpers {
...
+    void HelpersQmlWrapper::openFile(const QString &path) {
+        QStringList args;
+        args << path;
+
+#ifdef Q_OS_MAC
+        QProcess::startDetached("open", args);
+#endif
+
+#ifdef Q_OS_WIN
+        QProcess::startDetached("start", args);
```
ribtoks: Did you try it in windows? To my knowledge to open something on Windows with default application, let's say file `c:\image.jpg`, you just start a process with a command line `c:\image.jpg`, but not `start c:\image.jpg` (at least in C#)

velmafia: I [found](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/start) that `start` command is dedicated for this. Also we can use only file name, as you wrote, but I wanted to be consistent with `linux` and `osx` part.

To be honest, I don't tested this with xpisk, I check this command with `cmd` on windows system (I still haven't windows VM configured with development environment).

ribtoks: Don't worry, I can test it later on Windows

velmafia: ok, thx

PR #745, opened 2019-07-18
Issues/449 edit keyword suggestions

velmafia:
Problem with `EditKeywordDialog` focus still not solved - I marked this redundant `mouseClick` with comment `[HACK]`.


src/xpiks-tests/xpiks-tests-integration/spellchecksinglekeywordmodeltest.cpp
```diff
@@ -0,0 +1,71 @@
...
+    m_TestsApp.dispatch(QMLExtensions::UICommandID::SetupFixKeywordSpelling, wrongWord);
+
+    SpellCheck::SpellCheckSuggestionModel &spellSuggestor = m_TestsApp.getSpellSuggestionsModel();
+    int rowCount = spellSuggestor.rowCount();
+    VERIFY(rowCount == 1, "Spell suggestions are not set");
+
+    SpellCheck::SpellSuggestionsItem *suggestionsItem = spellSuggestor.getItem(0);
+    VERIFY(suggestionsItem->rowCount() > 0, "No spelling suggestion suggested");
+
+    spellSuggestor.getActionCommand(false)->execute();
```
ribtoks: If the test does not pass `VERIFY` will return from this function so "clear" by doing action command will not work obviously, because it will become unreachable code. In general this line should not be here and spell suggestor model should be cleaned up in `XpiksTestsApp::cleanup`. Please verify it is already cleaned up (e.g. by calling `resetModel()` or equivalent) or add missing code there. Either way this line should be removed from test.

src/xpiks-qt/Dialogs/EditKeywordDialog.qml
```diff
@@ -124,6 +171,7 @@ StaticDialogBase {
     }
 
     Component.onCompleted: {
+        dispatcher.dispatch(UICommand.SetupFixKeywordSpelling, previousKeyword)
```
ribtoks: This should be done before you do `Common.launchDialog(EditKeywordsDialog.qml)`. Otherwise you may see a visible UI lag or even worse, `spellCheckSuggestionModel.getSuggestionObject(0)` will always return `null` and you will not have any delegates. BTW, to check latest problem, can you try to dispatch this command on timer (e.g. 2s after the dialog is fully created) to see if it does not crash if it will create suggestions afterwards.

src/xpiks-qt/Services/SpellCheck/spellsuggestionsitem.h
```diff
@@ -98,6 +98,21 @@ namespace SpellCheck {
         bool m_ReplacementSucceeded;
     };
 
+    class SingleKeywordSpellSuggestions : public SpellSuggestionsItem {
+        Q_OBJECT
+    public:
+        SingleKeywordSpellSuggestions(const QString &word,
```
ribtoks: Just out of curiosity, in C++ it's possible to inherit a constructor with `using BaseClass::BaseClass`. I suppose it does not work if you have constructor overloads, like in this case?

src/xpiks-qt/Dialogs/EditKeywordDialog.qml
velmafia: In first version, I called `dispatch` before `launchDialog` but in this case dialog was faster than dispatch, this is reason why I implement whole `ListView`. 
So I had tested behavior when dialog start with empty model, and everything work fine (any crash doesn't appear). When I had implemented full `ListView` and I didn't observe any lags, I decide move `dispatch` here (still I don't see any lags), I think it will be more error proof than always remember that I need call `dispatcher.dispatch(UICommand.SetupFixKeywordSpelling, previousKeyword)` before `Common.launchDialog(EditKeywordsDialog.qml)`.

But If you think that will be better to call it before `launchDialog`, I can change this.

ribtoks: No, I don't want or think something, I only want it to work. So far I don't understand how it will work. Even without this change there's a possibility of race condition.

1. You call `Common.launchDialog(...)`, Qt starts rendering all the elements, spelling suggestor does not have any rows.
2. `Flow+Repeater` initialize, `.getSuggestionObject(0)` returns `nullptr`, `Flow` initializes with empty repeater.
3. spelling suggestor receives the suggestions, but UI is already rendered and bindings (to nullptr) are already in place

So how it is supposed to work in this case?

src/xpiks-qt/Dialogs/EditKeywordDialog.qml
```diff
@@ -103,6 +108,48 @@ StaticDialogBase {
...
+                    anchors.leftMargin: 5
+                    anchors.right: parent.right
+                    anchors.rightMargin: 5
+                    anchors.top: parent.top
+                    anchors.topMargin: 5
+                    spacing: 5
+                    focus: true
+
+                    Repeater {
+                        model: spellCheckSuggestionModel.getSuggestionObject(0)
```
ribtoks: OK, I just currently realized what have you implemented. There's a bug here: you should call `getSuggestionObject()` with real index, not `0`. If there will be `> 1` row in spell suggestor, you will render only first row multiple times. Just call it with `getSuggestionObject(suggestionsWrapper.delegateIndex)` (and add property `delegateIndex` to `suggestionsWrapper`).

src/xpiks-qt/Dialogs/EditKeywordDialog.qml
ribtoks: OK, now I understand what you have implemented. I left a comment above about index in delegate. Overall it should be completely OK as is.

src/xpiks-qt/Dialogs/EditKeywordDialog.qml
ribtoks: The fact that there must be only 1 row is a different story and should be checked in unit or integration tests. Here it should still render correctly.

velmafia: Missing `delegateIndex` - fixed.
My integration test already tested `rowCount == 1`, so I little change `VERIFY` message that we expect exactly one row.

src/xpiks-tests/xpiks-tests-integration/spellchecksinglekeywordmodeltest.cpp
velmafia: Done, `XpiksTestsApp::cleanup` already cleaned `SpellCheckSuggestionModel`.

src/xpiks-qt/Services/SpellCheck/spellsuggestionsitem.h
velmafia: If constructors from base class have the same signature (like in this case) then will be shadowed by constructor from derived class. So yes - in this case `using BaseClass::BaseClass` will not work.

ribtoks: No, that's not what I meant. If constructor (single) of Derived class has same signature and only thing it does - calls Base class constructor, in Derived class it can be removed and we can just add `using Base::Base;` declaration. However, I'm not sure we can use this trick here because base class has 2 constructors.

velmafia: Now I understand what you mean.
I think we can use this trick here (it should works: [example](https://coliru.stacked-crooked.com/a/394efb128aeab42b)), but for me explicitly defined constructor of derived class (like is now) is more clean and understandable.

PR #767, opened 2019-09-23
Optimize filtering logic

vleschuk:
This consists of two commits:

1. We introduce the PreparedFilter class and replace actual QString filter, filtering flags, etc in FilteredArtworksListModel with an instance of this class.

2. We separate actual filtering from filter parsing so that we parse filter only once when setting filtering string and after that use these results when performing actual search. We also optimize filter parsing by setting special search flags for reserved keywords like "x:image" so that when performing search we do not compare strings but just check flags.


ribtoks:
Looks almost perfect, thank you. Will wait until CI turns green.

ribtoks:
@vleschuk core tests fail on Travis CI and in AppVeyor. Can you please try to run them locally and fix the bugs? Code looks good, there should be some small typo.

Almost all Artwork Filter tests failed. It is convenient to debug them by commenting everything except them in `xpiks-tests-core/main.cpp` and running.

vleschuk:
Tests pass but the process exits with code 16. Can't reproduce this on master branch, so this PR needs some more debugging.

ribtoks:
@vleschuk tests didn't pass. Please check again. [ArtworkFilter tests](https://travis-ci.org/ribtoks/xpiks/jobs/588652396#L4739) failed and some other too. I believe if you will make artwork filter tests pass, others will also pass since they are dependent on same functionality.

vleschuk:
Yes, I see it now. Working on it.

ribtoks:
Merged to master. Thank you

Issue #753, opened 2019-08-07
Load plugins from json config instead of iterating directory


ribtoks:
part of #445 effort
dependent on #752 

This task is about loading plugins only from json file instead of iterating directory `/plugins/`. Json config from #752 should contain path to main plugin file and Xpiks should only load plugin from this path.

Currently all plugins that fail to load during directory iteration are put to smth like `/failed_plugins/` directory. After adding loading from json this must be deprecated and they should be kept where they are now. Only thing is they should be marked "errored" in the json config and not loaded again.

There's another task to display load errors in the UI.

Issue #752, opened 2019-08-07
Save installed plugins metadata to json


ribtoks:
part of #445 effort

Right now information about installed plugins is kept only in memory.

This task is about adding a json config (e.g. like `SettingsModel` has) to `PluginManager` class that will save useful plugin information in there. Useful plugin information is, but not limited to:
* filepath to installed plugin
* filepath to "raw" plugin package (from #751)
* installed version
* name
* enabled/disabled state
* sha1 sum of the main file (`.dll` etc.)
* last successful load date
* anything else?

Json file should be saved on every qualifying change (e.g. added/removed plugin, enabled/disabled plugin). This should be implemented same way as `SettingsModel` and others using `DelayedActionEntity` class and `justChanged()` API.

Also `PluginManager` should load json on load (or in some new method `PluginManager::initialize()` etc.) called from `XpiksApp::initialize` where all entities load json configs.

This task is only about adding such config. Actual usage of it (loading plugins comes as a separate task).

Also need to clear this config for integration tests in `::cleanup()` method between runs.

Plugin json config should be handled by a completely separate new class (like `PluginsConfig`) that will be a member of `PluginsManager`.

velmafia:
I take this task.

Issue #754, opened 2019-08-07
Display failed to load plugins in the UI as errored


ribtoks:
part of #445 effort
dependent on #753

When Xpiks loads plugins from json config and the plugin fails to load, Xpiks should mark the plugin as "errored" in the json config and not attempt to load such plugin on next run. However, this plugin should be visible in the UI and user must be able to uninstall it.

Add new property `errored` (or similar) to `PluginManager::roleNames()` and appropriate enum etc. If plugin errored, show it in UI same way as "Disabled" but disable all buttons instead of `"Remove"`.

Verify manually (or if needed add trivial test) that you can uninstall "errored" plugin without problems.

PR #786, opened 2019-11-21
Proposal for #752

velmafia:
wip: first implementation of plugins json config (#752)

The main idea is to store all values required in json config in one class (`PluginData`), this class should have method to export/import plugin data to json format, and due this `PluginConfig` may not care about what exactly fields should be stored in file. 

Next step what I want to do is small refactor of `PluginManager` - I want to use `PluginData` to store paths during plugin installation / loading (instead of passing it by argument as it's now). This should simplify next issue (#753), and `PluginManager` code.

@ribtoks  What do you think?


src/xpiks-qt/Plugins/pluginconfig.h
```diff
@@ -0,0 +1,59 @@
...
+
+#include "Common/delayedactionentity.h"
+#include "Helpers/localconfig.h"
+#include "Plugins/plugindata.h"
+
+namespace Plugins {
+
+    class PluginConfig :
+        public Common::DelayedActionEntity,
+        public QObject // do we need this?
```
ribtoks: Yes, you need to to work with `DelayedActionEntity` - it works on top of Qt Timers

src/xpiks-qt/Plugins/pluginconfig.h
```diff
@@ -0,0 +1,59 @@
...
+    protected:
+        virtual void doKillTimer(int timerId) override { this->killTimer(timerId); }
+        virtual int doStartTimer(int interval, Qt::TimerType timerType) override { return this->startTimer(interval, timerType); }
+        virtual void doOnTimer() override { saveFileConfig(); }
+        virtual void timerEvent(QTimerEvent *event) override { onQtTimer(event); }
+        virtual void callBaseTimer(QTimerEvent *event) override { QObject::timerEvent(event); }
+
+
+    private:
+    QMap<QString, PluginData> m_Plugins;
```
ribtoks: Why this is a map instead of a list?

src/xpiks-qt/Plugins/plugindata.h
```diff
@@ -0,0 +1,35 @@
...
+        QString m_Name;
+        QString m_Version;
+        QString m_FilePath;
+        QString m_PackagePath;
+        bool m_IsEnabled;
+        QString m_Sha1sum;
+        QString m_LastLoadDate;
+
+        QJsonObject toJson() const;
+        static PluginData fromJson(QJsonObject);
```
ribtoks: Maybe it can be a constructor?

src/xpiks-qt/Plugins/pluginmanager.cpp
```diff
@@ -68,7 +69,8 @@ namespace Plugins {
         m_MicrostockServices(microstockClients, requestsService),
         m_CurrentEditableSource(currentEditableSource),
         m_UIProvider(uiManager),
-        m_LastPluginID(0)
+        m_LastPluginID(0),
+        m_PluginConfig(m_Environment.path({}) + "/" + Constants::PLUGIND_FILENAME)
```
ribtoks: `m_Environment.path(Constants::PLUGIND_FILENAME)` will do it for you

src/xpiks-qt/Plugins/pluginmanager.cpp
```diff
@@ -405,6 +410,13 @@ namespace Plugins {
             const int pluginID = plugin->getPluginID();
             m_PluginsDict.insert(pluginID, plugin);
 
+            PluginData pd;
+            pd.m_Name = plugin->getPrettyName();
+            pd.m_Version = plugin->getVersionString();
+            pd.m_FilePath = plugin->getFilepath();
+            pd.m_IsEnabled = plugin->getIsEnabled();
+            m_PluginConfig.insertPlugin(pd);
```
ribtoks: What about `::addPlugin(plugin)` instead? Insert assumes you insert to some position

src/xpiks-qt/Helpers/constants.h
```diff
@@ -101,6 +101,7 @@ namespace Constants {
     const char SELECTED_LOCALE[] = "DEBUG_SELECTED_LOCALE";
     const char SELECTED_THEME_INDEX[] = "DEBUG_SELECTED_THEME_INDEX";
     const char FTP_STOCKS_FILENAME[] = "stocks_ftp.json";
+    const char PLUGIND_FILENAME[] = "plugins.json";
```
ribtoks: Typo - `PLUGIN` S `_FILENAME`

src/xpiks-qt/Plugins/pluginconfig.h
velmafia: Reason was that if I will need update some plugin, this map will be faster that iterating across list to find is this plugin already exist.

But this is still WIP, and I will probably change this.

src/xpiks-qt/Plugins/pluginconfig.h
velmafia: OK

src/xpiks-qt/Plugins/pluginconfig.h
ribtoks: I mean, we would need to have method `hasPlugin()` and all that, but this operation should not be frequent at all. It will be used once or twice per whole Xpiks process lifetime. Maybe no point in having a map?

ribtoks:
@velmafia good idea with `PluginData` and overall it seems OK. I left couple of super small comments inside.

src/xpiks-qt/Plugins/pluginconfig.h
ribtoks: Yes, but how frequently do you need to update a plugin? What comes to my mind is only Enable/Disable operations and we should be OK to make a loop through all of the plugins.

src/xpiks-qt/Plugins/plugindata.h
velmafia: If `QJsonObject` parsing will fail, the only way to exit from constructor is throw a exception - I don't like this solution.
This is also not perfect but with this function I can change return value (add `std::optional` for example), but this is WIP - so I not decide yet how to solve this.

src/xpiks-qt/Plugins/pluginmanager.cpp
velmafia: OK

src/xpiks-qt/Plugins/plugindata.h
ribtoks: Yes, I understand. But what is important here is that seeing this points (from your side) already suggests that current approach is incorrect (having it either as constructor or as static method).

It can be solved using "another level of indirection". Just parse it outside. Serialize/deserialize operations should belong to the list (vector, array). Then if you cannot parse next element, you don't add next list item. This ways it is clean. If you offload this decision to the element itself - then you face troubles that we are discussing.

Examples in Xpiks are `parseUploadInfos()`, `deserializeExportPlansArray()` for csv, `parsePresetArray()` - they all parse list and they don't have the problem we are discussing.

src/xpiks-qt/Plugins/plugininstaller.cpp
```diff
@@ -0,0 +1,156 @@
...
+#include <QUrl>
+#include <QVector>
+#include <QtDebug>
+
+#include "Common/logging.h"
+#include "Helpers/filehelpers.h"
+#include "Helpers/ziphelper.h"
+
+namespace {
+    QString sha1(const QString &filePath)
```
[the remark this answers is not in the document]
ribtoks: If it is used only in this place in whole app, then no. If it is used elsewhere then definitely a candidate for "filehelpers.h"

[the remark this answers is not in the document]
velmafia: I found one more usage of file checksum (`UpdatesCheckerWorker`), so moved to `filehelpers.h`

velmafia:
I added migration for plugin installed by old way to json config.

Now after installing plugins (or load old plugins) json config is generated, but still some functionalities are missing (for example remove PluginData when plugin is removed).

@ribtoks please take a look for my proposal (from architecture POV only, because code still need come cleanup).

And some questions:
1. What about multi-file plugins? (See my comment in code).
2. What is exactly id of plugin? Currently is file name, buy what about multi-file plugins? Plugin package filename will be enough? Unfortunately we can use "Pretty Name", because we can not get this value if plugins loading will fail.

ribtoks:
Hey, few comments before I look into the code.

1. There's no need to have a migration since plugins support is not a public feature. You can write code as nobody is using it.
2. Regarding missing functionality - I'm not sure I understand the logic. Let's say there's a `std::vector<PluginData>` (I would even call it metadata, not data). Every time you do some action (`justChanged()`) with plugins this vector is saved to a file. So plugin enablings/disablings and adding/removing automatically leads to a correct result - no need to handle it somehow differently.

src/xpiks-qt/Plugins/pluginconfig.h
```diff
@@ -0,0 +1,63 @@
...
+
+#include "Common/delayedactionentity.h"
+#include "Helpers/localconfig.h"
+#include "Plugins/plugindata.h"
+
+namespace Plugins {
+
+    class PluginConfig :
+        public Common::DelayedActionEntity,
+        public QObject
```
ribtoks: Whoops, Qt issue here: `QObject` inheritance _must_ be always first + you _must_ add `Q_OBJECT` macro in the beginning of the class

src/xpiks-qt/Plugins/pluginconfig.h
```diff
@@ -0,0 +1,63 @@
...
+        PluginConfig(const QString &filePath);
+
+        virtual ~PluginConfig() override;
+
+        void addOrUpdatePlugin(const PluginData &pd);
+        std::vector<PluginData> getPluginsData();
+
+        // if config doesn't exist, so this is probably first run of xpisk with json config
+        // [TODO] this check is enough for handling migration from previous version?
+        bool configExist() { return m_ConfigLoaded; }
```
ribtoks: Why do you need this API? `PluginConfig` (btw, why it is Plugin Config but not PluginS Config? you're handling multiple plugins here) on `read()` will return empty vector and that's it..

src/xpiks-qt/Plugins/pluginconfig.h
```diff
@@ -0,0 +1,63 @@
...
+    protected:
+        virtual void doKillTimer(int timerId) override { this->killTimer(timerId); }
+        virtual int doStartTimer(int interval, Qt::TimerType timerType) override { return this->startTimer(interval, timerType); }
+        virtual void doOnTimer() override { saveFileConfig(); }
+        virtual void timerEvent(QTimerEvent *event) override { onQtTimer(event); }
+        virtual void callBaseTimer(QTimerEvent *event) override { QObject::timerEvent(event); }
+
+    private:
+        Helpers::LocalConfig m_LocalConfig;
+        bool m_ConfigLoaded;
```
ribtoks: C++ will align structs so memory layout of your data on x86_64 will be
8 bytes of localconfig (let's assume it's just a pointer for simplicity)
1 byte of bool + 7 bytes padding
8 bytes of plugins vector

I prefer to put bool in the end so if you have few bools you can always combine them and save space. (I know, we're not coding for Arduino here, but still..)

src/xpiks-qt/Plugins/plugindata.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+#include "Plugins/plugindata.h"
+
+#include <QJsonArray>
+
+namespace Plugins {
+
+    QJsonObject PluginData::toJson() const {
+        QJsonObject plugin;
+        plugin.insert("name", m_Name);
```
ribtoks: These names are good candidates for extraction to some sort of constants since they will be used on reading too

src/xpiks-qt/Plugins/plugindata.h
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Plugins {
+
+    class PluginData {
+    public:
+        QString m_Name;
+        QString m_Version;
+        QVector<QString> m_PluginFiles;
+        QString m_PackagePath;
+        bool m_IsEnabled;
```
ribtoks: Can you pls put 2 bools in the end? You see, here you create 7+7 bytes of useless memory padding when you can have only 6 bytes of padding (if you put 2 bools in the end of this list, compiler will only pad 6 bytes on x86_64). 

I know there will not be enough plugins in the memory so by your paddings you will consume whole RAM, but this is just a useful habbit to have.

src/xpiks-qt/Plugins/pluginmanager.cpp
```diff
@@ -304,94 +335,34 @@ namespace Plugins {
...
-            packageDestination = QDir::cleanPath(m_PluginsPackagesPath + QChar('/') + QFileInfo(packagePath).fileName());
-            LOG_DEBUG << "Copying plugin package [" << packagePath << "] to [" << packageDestination  << "]";
-            if (!QFile::copy(packagePath, packageDestination)) {
-                LOG_WARNING << "Failed to copy plugin package to" << pluginDestination;
-                break;
-            }
-
-            success = true;
-        } while (false);
+        if (!installer.install()){
```
ribtoks: In my opinion here we have some naming conflict and below too (`installer.install()` just asks to get a parameter and `installer.uninstall()` too). If you would name it like `PluginPackage::install()` and `PluginPackage::remove()` in my opinion it might help with reading flow.

src/xpiks-qt/Plugins/pluginmanager.cpp
```diff
@@ -304,94 +335,34 @@ namespace Plugins {
...
+        m_PluginConfig.addOrUpdatePlugin(pluginData);
+        LOG_INFO << "Plugin" << pluginUrl << "added";
+        return true;
     }
 
-    bool PluginManager::doAddPlugin(const QString &filepath) {
+    bool PluginManager::doAddPlugin(PluginData &pd) {
         bool added = false;
+
+        auto filepath = pd.m_PluginFiles.front();  // [TODO] currently plugin has only one file
```
ribtoks: I'm not sure `.front()` is what we need. Plugin will always have only 1 main file (.dll, .so etc.) that will be loaded into the memory. Plugin may have other auxiliary files, but only 1 ~~executable~~ loadable. You may have a separate member of `PluginData` that will return "main file"

src/xpiks-qt/Plugins/pluginmanager.cpp
```diff
@@ -405,6 +376,11 @@ namespace Plugins {
             const int pluginID = plugin->getPluginID();
             m_PluginsDict.insert(pluginID, plugin);
 
+            pd.m_Name = plugin->getPrettyName();
+            pd.m_Version = plugin->getVersionString();
+            pd.m_IsEnabled = plugin->getIsEnabled();
+            pd.m_LastLoadDate = QDateTime::currentDateTimeUtc().toString();
```
ribtoks: Please use `.toString(Qt::ISODate);`

src/xpiks-qt/Plugins/pluginconfig.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+    PluginConfig::~PluginConfig() {
+        LOG_DEBUG << "#";
+        saveFileConfig();
+    }
+
+    void PluginConfig::addOrUpdatePlugin(const PluginData &plugin) {
+        LOG_DEBUG << plugin.m_Name;
+
+        const auto fn = plugin.m_PluginFiles.front();
```
ribtoks: See comment somewhere below about "main file"

src/xpiks-qt/Plugins/plugininstaller.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+#include "Common/logging.h"
+#include "Helpers/filehelpers.h"
+#include "Helpers/ziphelper.h"
+
+namespace Plugins {
+    PluginInstaller::PluginInstaller(const QUrl &filePath, const QString &dstDir) :
+        m_destinationDir(dstDir)
+    {
+        QFileInfo package(filePath.toLocalFile());
```
ribtoks: You're doing too many of smart actions in the constructor and now you need `m_Error`. Please use constructor to initialize variables here and do all this business logic with error handling in some other method.

src/xpiks-qt/Plugins/plugininstaller.cpp
```diff
@@ -0,0 +1,141 @@
...
+        }
+        else {
+            LOG_DEBUG << "Plugin " << m_PackagePath << " uninstalled completed";
+        }
+
+        return !m_Error;
+    }
+
+    PluginData PluginInstaller::getPluginData() {
+        PluginData pd;
```
ribtoks: I'd say that we may create this `PluginData` internally and then only return `const &` in this method.

ribtoks:
I must say that I like where this is moving to, but now it looks like you started to put too much into 1 change..

src/xpiks-qt/Plugins/pluginmanager.cpp
```diff
@@ -116,26 +103,66 @@ namespace Plugins {
...
-            auto pluginWrapper = loadPlugin(pluginFullPath);
-            if (pluginWrapper) {
-                loadedPlugins.push_back(pluginWrapper);
-                Q_ASSERT(!pluginsDict.contains(pluginWrapper->getPluginID()));
-                pluginsDict.insert(pluginWrapper->getPluginID(), pluginWrapper);
-            } else {
-                processInvalidFile(fileName, pluginFullPath);
+                const QString pluginFullPath = pluginsDir.absoluteFilePath(fileName);
+
+                // migration plugins data to json
```
ribtoks: As I already mentioned, there's no need to run any migration. Nobody except of 1 company in Germany uses Xpiks plugins. This is not yet a public feature so you can skip migrations and just work as is.

src/xpiks-qt/Plugins/pluginmanager.cpp
```diff
@@ -116,26 +103,66 @@ namespace Plugins {
...
+                }
+
+                auto pluginWrapper = loadPlugin(pluginFullPath);
+                if (pluginWrapper) {
+                    loadedPlugins.push_back(pluginWrapper);
+                    Q_ASSERT(!pluginsDict.contains(pluginWrapper->getPluginID()));
+                    pluginsDict.insert(pluginWrapper->getPluginID(), pluginWrapper);
+
+                    pd.m_IsFailed = false;
+                    pd.m_Name = pluginWrapper->getPrettyName();
```
ribtoks: I think all these assignments should happen inside some method of `PluginData` like `PluginData::load(pluginWrapper)` or anything along those lines. it's better to keep them in 1 place and especially not to pollute code in plugins manager. This "install()" function was one of the very few functions in Xpiks that were longer than 1 screen of code so we should strive to keep it minimal..

src/xpiks-qt/Plugins/pluginmanager.cpp
```diff
@@ -116,26 +103,66 @@ namespace Plugins {
...
+                    pd.m_LastLoadDate = QDateTime::currentDateTimeUtc().toString();
+                } else {
+                    pd.m_IsFailed = true;
+                }
+
+                m_PluginConfig.addOrUpdatePlugin(pd);
+            }
+        }
+        else {
+            for (auto &pd : m_PluginConfig.getPluginsData()) {
```
ribtoks: I already left a comment like this, but just to emphasize: we don't care about migrations, plugins will only become public feature in version 1.6

ribtoks:
Sorry, I forgot to answer a question
> What is exactly id of plugin? Currently is file name, buy what about multi-file plugins? Plugin package filename will be enough? Unfortunately we can use "Pretty Name", because we can not get this value if plugins loading will fail.

So ID of the plugin is what we decide. It can be a integer or a string, it does not matter. I think it's more convenient to have an integer - like when you install a new plugin, you assign a unique ID (autoincrement) and always use it through the config. If it's not possible right now, we can use the filename of the main loadable file.

In multi-files plugin there still will only be a single loadable file (.dll/.so/.dylib) so there's nothing to worry about.

src/xpiks-qt/Plugins/pluginmanager.cpp
velmafia: changed

src/xpiks-qt/Plugins/pluginmanager.cpp
velmafia: ok

src/xpiks-qt/Plugins/pluginmanager.cpp
velmafia: ok

src/xpiks-qt/Plugins/plugindata.cpp
ribtoks: If it's used only in this file, will be ok.

src/xpiks-qt/Plugins/plugindata.h
velmafia: Changed in the newest commit, something like this is better?

src/xpiks-qt/Plugins/packageinstaller.cpp
```diff
@@ -0,0 +1,154 @@
...
+        m_DestinationDir(dstDir)
+    {
+        Helpers::ensureDirectoryExists(m_DestinationDir);
+    }
+
+    PackageInstaller::PackageInstaller(const PluginMetadata &metadata) :
+        m_PluginMetadata(metadata)
+    {}
+
+    bool PackageInstaller::install() {
```
ribtoks: Hi. I already mentioned this naming issue, but I'd like to point it out again. By having `PackageInstaller::install()` API you make a convention that you `install()` a `PackageInstaller` instance which is incorrect. A better way would be to `install()` or `uninstall()` a `PluginPackage`. If it was a `PluginManager::install(PluginPackage)` it would be even better semantically, but not in our case practically.

In OOP we should name clases after what they **are**, not after what they **do**. Since we don't actually have a plugin installer, but we have a plugin archive or package, I propose to go after this name.

src/xpiks-qt/Plugins/packageinstaller.cpp
```diff
@@ -0,0 +1,154 @@
...
+            LOG_WARNING << "Failed to get package " << m_OriginPath;
+            return false;
+        }
+
+        if (!unpackPackage()) {
+            LOG_WARNING << "Failed to unpack plugin " << m_PluginMetadata.m_PackagePath;
+            return false;
+        }
+
+        LOG_DEBUG << "Plugin " << m_OriginPath << " installed completed";
```
ribtoks: nit: no need to add space in " installed ..." string, Qt does it for you

src/xpiks-qt/Plugins/packageinstaller.cpp
```diff
@@ -0,0 +1,154 @@
...
+        return true;
+
+    }
+
+    bool PackageInstaller::uninstall() {
+        LOG_DEBUG << "Uninstalling " << m_PluginMetadata.m_PackagePath << " plugin";
```
ribtoks: No need to add space, Qt does it for you

src/xpiks-qt/Plugins/packageinstaller.cpp
```diff
@@ -0,0 +1,154 @@
...
+            LOG_DEBUG << "Plugin " << m_PluginMetadata.m_PackagePath << " uninstalled completed";
+        }
+        else {
+            LOG_WARNING << "Package " << m_PluginMetadata.m_PackagePath << "uninstalled with error, see previous logs";
+        }
+
+        return ret;
+    }
+
+    const PluginMetadata &PackageInstaller::getPluginMetadata() {
```
ribtoks: Method should be `const`

src/xpiks-qt/Plugins/packageinstaller.cpp
```diff
@@ -0,0 +1,154 @@
...
+            LOG_DEBUG << "Plugin contains " << packageFiles.size() << "file(s)";
+        }
+        else {
+            LOG_WARNING << "Plugin is empty! Nothing to install";
+            return false;
+        }
+
+        for (const auto &file : packageFiles) {
+            QFileInfo srcFile(file);
+            QString dst = QDir::cleanPath(m_DestinationDir + QChar('/') + srcFile.fileName());
```
ribtoks: Please use `QDir::separator()`

src/xpiks-qt/Plugins/pluginconfig.cpp
```diff
@@ -0,0 +1,128 @@
...
+    bool PluginConfig::loadFileConfig() {
+        LOG_DEBUG << "#";
+        auto config = m_LocalConfig.readConfig();
+
+        if (config.isEmpty()) {
+            LOG_DEBUG << "Plugin config file is empty.";
+            return false;
+        }
+
+        m_Plugins.clear();
```
ribtoks: I propose a better pattern:
```
decltype(m_Plugins) plugins;
// --> here parse plugins
m_Plugins.swap(plugins);
```

Otherwise you during this method you will have `m_Plugins` in inconsistent state of for some reason we will use threading here.

src/xpiks-qt/Plugins/pluginconfig.cpp
```diff
@@ -0,0 +1,128 @@
...
+        if (config.isArray()) {
+            for (const auto elem : config.array()) {
+                auto plugin = tryParsePluginMetadata(elem.toObject());
```
ribtoks: Usually semantics of `tryDoSomething()` method is like this:
`bool tryDoSomething(Type1 param1, ..., ResultType &result);`

src/xpiks-qt/Plugins/pluginmanager.cpp
```diff
@@ -267,12 +258,20 @@ namespace Plugins {
...
 
-            const QString fullPath = wrapper->getFilepath();
-            Helpers::removeFile(fullPath);
-
-            const QString filename = QFileInfo(fullPath).completeBaseName();
-            const QString packageFile = QDir::cleanPath(m_PluginsPackagesPath + "/" + filename + Constants::XPIKS_PLUGIN_EXTENSION);
-            Helpers::removeFile(packageFile);
+            auto pluginMD = m_PluginConfig.getPluginByMainFile(wrapper->getFilepath());
+            if (pluginMD) {
+                PackageInstaller pluginPackage(*pluginMD);
```
ribtoks: Would be more consistent if it was `PluginPackage pluginPackage(*pluginMD)`. How do you think?

src/xpiks-qt/Plugins/pluginmetadata.cpp
```diff
@@ -0,0 +1,115 @@
...
+            if(!additionalFilesValue.isArray()) { break; }
+            else {
+                for (const auto file : additionalFilesValue.toArray()) {
+                    if (file.isString()) {
+                        additionalFiles.push_back(file.toString());
+                    }
+                }
+            }
+
+            ret = std::make_unique<PluginMetadata>();
```
ribtoks: Let's assign values right after validation (on top), otherwise it is very hard to track that you validate or assign all values

src/xpiks-qt/Plugins/pluginmetadata.h
```diff
@@ -0,0 +1,40 @@
...
+        QString m_PackagePath;
+        QString m_PackageSha1;
+        QString m_LastLoadDate;
+        bool m_IsEnabled;
+        bool m_IsFailed;
+
+        QJsonObject toJson() const;
+    };
+
+    std::unique_ptr<PluginMetadata> tryParsePluginMetadata(const QJsonObject &json);
```
ribtoks: I propose to have this as a member of `PluginMetadata`.
Method will be `bool tryParse(const QJsonObject &json)`. This way you will not enforce user of your API to use `std::unique_ptr`. And you will comply will `tryDoSomething()` semantics in almost all other frameworks

src/xpiks-common/xpiks-common.pri
```diff
@@ -13,6 +13,7 @@ linux {
 DEPS_DIR = $$PWD/../xpiks-qt/deps
 
 DEFINES += DEPS_DIR=$${DEPS_DIR}
+DEFINES += SHLIB_EXTENSION=$${QMAKE_EXTENSION_SHLIB}
```
ribtoks: Cool trick!

src/xpiks-qt/Plugins/packageinstaller.cpp
```diff
@@ -0,0 +1,154 @@
...
+#include "Common/logging.h"
+#include "Helpers/filehelpers.h"
+#include "Helpers/ziphelper.h"
+
+namespace Plugins {
+    PackageInstaller::PackageInstaller(const QUrl &originPath, const QString &dstDir) :
+        m_OriginPath(originPath.toLocalFile()),
+        m_DestinationDir(dstDir)
+    {
+        Helpers::ensureDirectoryExists(m_DestinationDir);
```
ribtoks: Can we please move it out of constructor? Side-effects of object constructor are not very healthy on the long run in any project.

src/xpiks-qt/Plugins/packageinstaller.cpp
```diff
@@ -0,0 +1,154 @@
...
+        if (!Helpers::removeFile(m_PluginMetadata.m_PackagePath)) {
+            ret = false;
+        }
+
+        if (ret) {
+            LOG_DEBUG << "Plugin " << m_PluginMetadata.m_PackagePath << " uninstalled completed";
+        }
+        else {
+            LOG_WARNING << "Package " << m_PluginMetadata.m_PackagePath << "uninstalled with error, see previous logs";
+        }
```
ribtoks: I propose to change this pattern to this:
```
bool succeed = false;
do {
  if (NOT condition1) {
    LOG_WARNING << "description 1";
    break;
  }
  if (NOT condition2) {
    //...
    break;
  }

  success = true;
}
```

src/xpiks-qt/Plugins/packageinstaller.cpp
```diff
@@ -0,0 +1,154 @@
...
+    bool PackageInstaller::copyPackage() {
+        QFileInfo package(m_OriginPath);
+
+        if (!package.exists()) {
+            LOG_WARNING << "Package not found:" << m_OriginPath;
+            return false;
+        }
+
+        QString packageDst = QDir::cleanPath(m_DestinationDir + QChar('/') + package.fileName());
+        if (QFile::exists(packageDst)) {
```
ribtoks: `QFileInfo(packageDst)::exists()` is a more lightweight alternative

src/xpiks-qt/Plugins/pluginconfig.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+        justChanged();
+    }
+
+    void PluginConfig::remove(const QString &mainFilePath) {
+        auto it = std::find_if(m_Plugins.begin(), m_Plugins.end(),
+                                        [&mainFilePath](const auto &pd) {return mainFilePath == pd.m_MainFile;} );
+        if (it != m_Plugins.end()) { m_Plugins.erase(it); }
+
+        justChanged();
```
ribtoks: Please move it to `if (it != m_Plugins.end())` since if we don't do any operations, we don't need to resave the list

src/xpiks-qt/Plugins/pluginconfig.cpp
```diff
@@ -0,0 +1,128 @@
...
+        std::unique_ptr<PluginMetadata> ret;
+
+        auto it = std::find_if(m_Plugins.begin(), m_Plugins.end(),
+                                        [&mf](const auto &pd) {return mf == pd.m_MainFile;} );
+        if (it != m_Plugins.end()) { ret = std::make_unique<PluginMetadata>(*it); }
+
+        return ret;
+    }
+
+    bool PluginConfig::isPackageInstalled(const QString &package) {
```
ribtoks: Just a question:

- where will we do more thorough checks like if the file exists on disk, hashsums etc.? (here we only check if it's in the vector)

ribtoks:
RE: check_style and friends. Please do NOT include even more changes in this PR. Let it pass green and change this define vs const thing later. This PR is already too big for me to digest, I need to jump back and forth frequently.

ribtoks:
RE: cppcheck. You can just remove it at all but please don't do it as part of this PR. You can submit separate PR and I will approve it before you submit this one.

src/xpiks-qt/Plugins/pluginmetadata.h
velmafia: done

src/xpiks-qt/Plugins/pluginmetadata.cpp
velmafia: outdated, after [this](https://github.com/ribtoks/xpiks/pull/786/files#r352264559)

src/xpiks-qt/Plugins/pluginmanager.cpp
velmafia: done

src/xpiks-qt/Plugins/pluginconfig.cpp
velmafia: outdated, after [this](https://github.com/ribtoks/xpiks/pull/786/files#r352264559)

ribtoks:
@velmafia I'm not exactly sure which tests do you think about. Serialize/deserialize?
RE:`PluginInstaller` - if you have a better suggestion, please tell me. I think this has has it's cons and if you don't like mine then please suggest a better one.

velmafia:
Regarding tests, I want to cover:
- `PluginMetadata` - serialize/deserialize
- `PluginPackage` - install/uninstall (destination is passed as argument, so we can `QTemporaryDir` as location for test)
- extend `LoadPluginBasicTest` for testing changes related with this new json

----
About `PluginInstaller` / `PluginPackage`, I don't have better name, but `PluginPackage` is associated me with simply plugin (file with `.xpiskplugin`).

I have some analogy to Debian system package:
- `.deb` - `.xpiksplugins`
- `apt-get` - "Package Manager", but this class is to simply to be `PackageManager`
- `dpkg` - it's fits better, but I don't have better name as simply "installer".

On the other hand, `PackageManager` can be good name if Plugins will be extended in feature (for example if we add remote repository with plugins).

ribtoks:
RE: tests. Yes, it makes sense to add them in this PR then.
RE: naming. 
Look, making analogies with package managers is bad because we are not creating a package manager but writing code in Object-Oriented Style. 

If there would be a class `PackageInstaller`, it's method `install()` would need to take an address of a package to install and you would instantiate it like `PackageInstaller(packagesRoot).install(packagePath)`. Then it will be logical. 

In your case `PackageInstaller(packagePath).install()` is less ideal in my opinion. 

And, yes, `PluginPackage` should be associated with `.xpiksplugin`. What is the problem with it? (i'm trying to understand)

velmafia:
About `PluginPackage` name, with my understanding this better fits to `.xpiksplugin` file, but what we have here, is something more like `PackageWrapper`, or something similar.  
But OK, let it be as is.

velmafia:
Almost done.

TODO list:
- [ ] fix CI for `osx` and `windows`
- [ ] add "cleanup" for plugins files for integration tests
- [ ] extend `LoadPluginBasicTest` (?)

src/xpiks-qt/Plugins/pluginmanager.cpp
```diff
@@ -267,12 +256,20 @@ namespace Plugins {
...
-            const QString packageFile = QDir::cleanPath(m_PluginsPackagesPath + "/" + filename + Constants::XPIKS_PLUGIN_EXTENSION);
-            Helpers::removeFile(packageFile);
+            auto pluginMD = m_PluginConfig.getPluginByMainFile(wrapper->getFilepath());
+            if (pluginMD) {
+                PluginPackage pluginPackage(*pluginMD);
+                if (!pluginPackage.uninstall()) {
+                    LOG_WARNING << "Some problem occurred during remove files for plugin" << pluginMD->m_Name;
+                    pluginMD->m_IsFailed = true;
+                    // don't remove plugin metadata, only mark it as failed for manually cleanup
+                    m_PluginConfig.addOrUpdate(*pluginMD);
```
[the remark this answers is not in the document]
velmafia: From one POV you are right, what more we can do if removed failed. But from other hand if plugin will be not fully removed, and we don't inform user about this, we probably will block re-installation or installation new version of this plugin, without any information for user. If we mark plugin as failed, we give some information for user that this plugin if broken (after #754).

I think we need some specification work with this problem (and generally with plugins).

src/xpiks-qt/Plugins/pluginmanager.cpp
```diff
@@ -405,28 +331,17 @@ namespace Plugins {
             const int pluginID = plugin->getPluginID();
             m_PluginsDict.insert(pluginID, plugin);
 
+            pluginMD.m_Name = plugin->getPrettyName();
+            pluginMD.m_Version = plugin->getVersionString();
+            pluginMD.m_IsEnabled = plugin->getIsEnabled();
+            pluginMD.m_LastLoadDate = QDateTime::currentDateTimeUtc().toString(Qt::ISODate);
```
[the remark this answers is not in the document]
velmafia: done

src/xpiks-qt/Plugins/pluginconfig.h
```diff
@@ -0,0 +1,67 @@
...
+    public:
+        PluginConfig(const QString &filePath);
+        virtual ~PluginConfig() override;
+
+    public:
+        void addOrUpdate(const PluginMetadata &plugin);
+        void remove(const QString &mainFilePath);
+
+        const std::vector<PluginMetadata> &getPluginsMetadata();
+        std::unique_ptr<PluginMetadata> getPluginByMainFile(const QString &mf);
```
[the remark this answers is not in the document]
ribtoks: `bool tryDoX(int param, int &result)` is not specifically a C-style API, it's more like a convention how not to use exceptions. I totally agree about your point for `::at(size_t)`, this is an implementation detail. So then this API should return a reference to the element instead of index.

What do you think about this:
- instead of `std::vector<PluginMetadata>` you use `std::vector<std::shared_ptr<PluginMetadata>>`
- and you change API to `bool tryFindPlugin(const QString &path, std::shared_ptr<PluginMetadata> &out)`
?

I partially agree about C-likeness of the `tryGetX()` API, but also want to point out that comparison with NULL is bad too and we should eliminate it as much as possible (especially in new code). `std::optional` is the same `NULL`, just with a nice name for non-nullable types and light taste of functional language (yay, let's add more languages inside c++).

src/xpiks-qt/Plugins/pluginmanager.cpp
```diff
@@ -116,27 +101,30 @@ namespace Plugins {
...
-            const QString pluginFullPath = pluginsDir.absoluteFilePath(fileName);
-            auto pluginWrapper = loadPlugin(pluginFullPath);
+            auto pluginWrapper = loadPlugin(pluginMD.m_MainFile);
             if (pluginWrapper) {
                 loadedPlugins.push_back(pluginWrapper);
                 Q_ASSERT(!pluginsDict.contains(pluginWrapper->getPluginID()));
                 pluginsDict.insert(pluginWrapper->getPluginID(), pluginWrapper);
+
+                pluginMD.m_LastLoadDate = QDateTime::currentDateTimeUtc().toString();
+                pluginMD.m_IsFailed = false;
```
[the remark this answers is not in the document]
ribtoks: No, I mean set `m_IsFailed` to `false` always as a matter of fact of initializing booleans to `false` in constructor of `PluginMetadata`. And set it to `true` only if loading fails

src/xpiks-qt/Plugins/pluginmanager.cpp
```diff
@@ -267,12 +256,20 @@ namespace Plugins {
             int removedCount = m_PluginsDict.remove(pluginID);
             Q_ASSERT(removedCount == 1);
 
-            const QString fullPath = wrapper->getFilepath();
-            Helpers::removeFile(fullPath);
-
-            const QString filename = QFileInfo(fullPath).completeBaseName();
-            const QString packageFile = QDir::cleanPath(m_PluginsPackagesPath + "/" + filename + Constants::XPIKS_PLUGIN_EXTENSION);
-            Helpers::removeFile(packageFile);
+            auto pluginMD = m_PluginConfig.getPluginByMainFile(wrapper->getFilepath());
```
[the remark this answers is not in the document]
ribtoks: C++17 is not enabled just because I had to support 32-bit build on Windows where I used Visual Studio 2015 without C++17 support. But now I decided to drop it, so theoretically there's no limitation.

RE: this C-thing topic, replied in different thread, let's keep it there

src/xpiks-qt/Plugins/pluginmanager.cpp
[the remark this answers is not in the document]
ribtoks: Well, plugin installation should not fail just because there're leftovers of previous plugin in some path. Before that would be a problem, but now with your feature we save path to the plugin in the config. We can install plugin to **any** directory we want as long as we know where to load it from. If the "default target directory" is busy, we should try to find "directory-1" etc. or just add some sort of a timestamp to directory name.

So I think we should do this:
- if we fail to install plugin we never include it to the config
- if we fail to remove the plugin we do not remove it from the config, but mark it as "failed to remove"
- on the next startup of Xpiks after the initial plugin load if we find "failed to remove" plugin, we retry to remove it
- if we remove a plugin successfully, we remove it from config

The only reasons for "failed to remove" could be that
- plugin failed to unload from memory (some dll hell things)
- file is locked because some process (maybe even process started by xpiks) has opened this file

Both of these issues should go away next time when we do not load the plugin.

src/xpiks-qt/Plugins/pluginpackage.cpp
```diff
@@ -0,0 +1,154 @@
...
+            if (!Helpers::removeFile(file)) {
+                ret = false;
+            }
+        }
+
+        if (!Helpers::removeFile(m_PluginMetadata.m_MainFile)) {
+            ret = false;
+        }
+
+        if (!Helpers::removeFile(m_PluginMetadata.m_PackagePath)) {
```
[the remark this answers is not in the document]
ribtoks: So as discussed in other comments, we should:
- try to cleanup first
- if failed, find a "substitution" candidate (more on this below)

We anyway only save a path to the file to load so the path can be any. I propose the following trick:

Each plugin should have a unique autoincrement ID (just int32). When reading config, you know max existing ID and each next plugin will have ID+1, ID+2 and so on. If there're no plugins, ID == 1 (I propose to reserve `0` for errors).

Then, when installing a plugin when there're existing files AND you fail to remove them, just add this plugin's ID to directory name, which is guaranteed to be unique, because you keep track of it. If such path exists, just keep incrementing ID until it doesn't exist.

In general case you don't use IDs, but you save them in `PluginMetadata` for tracking purposes.

What do you think?

src/xpiks-qt/Plugins/pluginconfig.h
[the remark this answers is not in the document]
velmafia: Regarding `std::vector<std::shared_ptr<PluginMetadata>>` I don't like this solution. When I see `std::shared_ptr` my first thought is that we shared ownership of this object, but in this case we don't do that. Ownership of this objects is well know `PluginConfig` is only owner.
Maybe my proposal with `std::unique_ptr` is overkill, now I think that I can simply return raw pointer, because ownership is not changed, but it doesn't matter...

If you prefer `bool tryDoX(int param, int &result)` maybe I can use this style and simply override `result` by copy of the found object (before I create "empty" `result`)? This is only remove - very rare functionality, so will be not impact for performance runtime.

---
> but also want to point out that comparison with NULL is bad too 

What exaclly do you mean? I'm curious, for me this is one of the "cleanest" way to solve this kind of problems (especially `std::optional`).

src/xpiks-qt/Plugins/pluginmanager.cpp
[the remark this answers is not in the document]
velmafia: answered in previous thread.
If we update to `c++17` in future, can I solve this kind of problems with `std::optional`? Question is not related with this PR, but for next PR's

src/xpiks-qt/Plugins/pluginconfig.h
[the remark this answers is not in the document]
ribtoks: Part1 - about optional.

**Overall answer below is all about semantics.** Semantics should be 100% clear even if it looks worse than "cooler" semantics. Maybe this is just me, but through my career in general now I strive to write as simple code as it is possible (and trying to rewrite complex parts of Xpiks). Because when in a year you read a piece of code and you don't have to stop at each line to understand what is written there - this is what is the most valuable.

First point is that `std::optional<>` is the same as [System.Nullable<T>](https://docs.microsoft.com/en-us/dotnet/api/system.nullable-1?view=netframework-4.8) from C# or just any class in Java. When your return a `std::optional<>` somebody needs to ~~check for null~~ check if optional has value (this is 100% the same as checking in C if value is NULL, just with more syntax sugar and only somewhat added safety).

The problem I see with it that semantics is screwed (both with nulls and with optional). Your method is called `getPluginByMainFile()` which is now incorrect name because it can return nothing instead of a plugin. So correct name for the user of your API would be `getPluginOrNothingByMainFile()`. Then it will be clear that you need to check the result if it's nothing or something. I'm exaggerating a little bit to make a point. 

Again, I'm nowhere a fan of `bool tryGetX(&x)` thing, but when you read this code, you have 0 questions about what it does and how do you need to use it.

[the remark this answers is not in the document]
ribtoks: Part 2 - about this particular place.

Way number 1.
I can suggest: never remove from vector. Add a new `bool m_IsRemoved` to the struct and keep it in vector forever. It will barely consume any additional memory. In the serialization in `justChanged()` you just skip elements that are removed and that's all. This way you will not need to worry about validity of the raw references that you return. e.g. you returned a reference to the plugin metadata in the end of the vector and then you removed 1st element and this reference is invalid and you have a crash.

Way number 2.
I didn't add this comment before because this PR is already super huge, but your `PluginInstaller` or `PluginPackage` constructors are kind of inconsistent. They accept too different data and they just ask to be split into 2 classes: `Installed-Plugin-Package` and `To-Be-Installed-Plugin-Package` or to consolidate the members and to have 1 thing: either metadata or `QString` with path. The whole reason why you return `PluginMetadata` from this method `getPluginByMainFile()` is to use this plugin metadata as parameter to constructor of `PluginPackage`.  So if you would change the semantics of `PluginPackage` constructor, you will solve this problem too.

P.S. This is a perfectionism and I'm afraid at some point we will need to leave this nice discussion and just move on with whatever solution is working right now. Let's say if you will not figure out nice way to solve this in 30 minutes, just leave it as is. OK?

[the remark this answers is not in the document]
velmafia: Because I don't have better idea how to change this API in nice way, I leave it as is.

But I also don't like this method, so I will keep in mind that this API is should be changes (or removed, it is used only in one place, so maybe I will find better solution).

src/xpiks-qt/Plugins/pluginmanager.cpp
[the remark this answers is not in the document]
velmafia: Ok, I can do this in that way, but can it be in next issue/PR? Now this PR is to big also for me.

src/xpiks-qt/Plugins/pluginpackage.cpp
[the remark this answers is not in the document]
velmafia: I think we can connect this task with your [proposition in other comment](https://github.com/ribtoks/xpiks/pull/786/files/6373d5bc0f034ad1a4057bbf39804ebc367e5b0e?file-filters%5B%5D=.cpp&file-filters%5B%5D=.pri&file-filters%5B%5D=.pro&file-filters%5B%5D=.qml&file-filters%5B%5D=.sh&file-filters%5B%5D=dotfile#r352948287), but maybe in another issue/PR.

src/xpiks-qt/Plugins/pluginmetadata.cpp
```diff
@@ -0,0 +1,141 @@
...
+        if (success) {
+            LOG_DEBUG << "Parsing plugin" << m_Name << "from json file - completed";
+        }
+        else {
+            LOG_DEBUG << "Parsing json element" << json << "failed, PluginMetadata is corrupted";
+        }
+        return success;
+    }
+
+    bool PluginMetadata::operator==(const PluginMetadata &other) const {
```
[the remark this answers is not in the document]
ribtoks: Probably `m_IsEnabled` should not be compared since it's the same plugin, just disabled.

src/xpiks-qt/Plugins/pluginmanager.cpp
```diff
@@ -289,109 +282,38 @@ namespace Plugins {
...
-    bool PluginManager::preparePackage(const QString &packagePath,
-                                       const QString &tmpPath,
-                                       QString &destination) {
-        bool success = false;
-        QString pluginDestination;
-        QString packageDestination;
-
-        do {
-            auto packageFiles = Helpers::unzipFiles(packagePath, tmpPath);
-            // TODO: handle plugin packages with more than 1 file
```
ribtoks: Please do not remove these TODOs unless you really solve the problem that they are describing. I'm using these todos as a task-management system described here: https://intmaker.com/2019/how-to-return-to-flow/

src/xpiks-tests/xpiks-tests-integration/loadpluginbasictest.cpp
```diff
@@ -26,10 +30,46 @@ int LoadPluginBasicTest::doTest() {
 
     Plugins::PluginManager &pluginManager = m_TestsApp.getPluginManager();
     bool success = pluginManager.installPlugin(QUrl::fromLocalFile(pluginPath));
-    VERIFY(success, "Failed to install plugin");
+    VERIFY(success, "Failed to install plugin")
+    QTest::qWait(CONFIG_SAVING_INTERVAL + 500); // wait for DelayedActionEntity
```
velmafia: This line extend test by 3,5s, we should wait also after `removePlugin(..)`, so finally test has 7s sleep. Do we can somehow trigger `DelayedActionEntity` faster?

ribtoks: Sorry, a little bit busy atm.

What about `#ifdef CORE_TESTS` and make save interval 1 sec for plugins config?

velmafia: I thinking about some smarter solution, but this should also works.
`#ifdef` added

velmafia:
One more thing:
In issue description you requested:
> Also need to clear this config for integration tests in ::cleanup() method between runs.

Can I do this in separate PR/issue? Currently leaving this config file don't have side effect, and this PR is really to big.

src/xpiks-tests/xpiks-tests-core/xpiks-tests-core.pro
```diff
@@ -78,9 +93,12 @@ SOURCES += main.cpp \
     ../../xpiks-qt/Artworks/vectorartwork.cpp \
     ../../xpiks-qt/Filesystem/indexedfilescollection.cpp \
     ../../xpiks-qt/Services/Maintenance/filescleanup.cpp \
+    ../../xpiks-tests/xpiks-tests-integration/testshelpers.cpp \
     encryption_tests.cpp \
     indicestoranges_tests.cpp \
     addcommand_tests.cpp \
+    pluginmetadatatests.cpp \
```
ribtoks: Can I ask you to follow the core tests pattern? All files are `<name>_tests.cpp`?

ribtoks:
@velmafia regarding config cleanup: created #792 for tracking purposes

src/xpiks-qt/Plugins/pluginmanager.cpp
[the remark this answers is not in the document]
velmafia: created separate issue for that: #794

src/xpiks-tests/xpiks-tests-core/xpiks-tests-core.pro
velmafia: done

velmafia:
My tests is failing for windows target..

- [2 core tests](https://ci.appveyor.com/project/Ribtoks/xpiks/builds/29373978/job/ebs9iyqn9u68a5no) have a problem with `QTemporaryDir` probably problem is related with [QT bugs](https://bugreports.qt.io/browse/QTBUG-64643?focusedCommentId=379790&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel)
- `LoadPluginBasicTest` is [also failing for windows](https://ci.appveyor.com/project/Ribtoks/xpiks/builds/29373978/job/sulmk20mh9xu5pd9), but to be honest - I don't know why, and I don't have working windows to test it.

@ribtoks can you test if this tests is working in your environments?

ribtoks:
@velmafia Sure, I'll take a look tomorrow. I understand about `QTemporaryDir`, but `LoadPluginBasicTest` used to work before..

ribtoks:
@velmafia Just a tip: there're full appveyor logs in Artifacts that can help you
image
It looks like you removed that code that unloads dll from memory first... But I'll debug tomorrow this
```
21:59:13.553 info T#2604 Models::UIManager::removePluginTab - Plugin's tab 42 removed
21:59:13.553 info T#2604 Plugins::PluginWrapper::removePlugin - "HelloWorldPlugin" "v0.0.1"
21:59:13.553 info T#2604 Plugins::PluginPackage::uninstall - Uninstalling "C:/Users/appveyor/AppData/Roaming/Xpiks/Xpiks_integration-tests/HEAD/session-06122019-215843/plugins/xpiks-helloworld-plugin0.xpiksplugin" plugin
21:59:13.553 debug T#2604 Helpers::removeFile - Attempting to remove file "C:/Users/appveyor/AppData/Roaming/Xpiks/Xpiks_integration-tests/HEAD/session-06122019-215843/plugins/xpiks-helloworld-plugin0.dll"
21:59:13.554 info T#2604 Helpers::removeFile - FAKE_WARNING Removing "C:/Users/appveyor/AppData/Roaming/Xpiks/Xpiks_integration-tests/HEAD/session-06122019-215843/plugins/xpiks-helloworld-plugin0.dll" failed, error: "Access is denied."
21:59:13.556 info T#2604 Helpers::removeFile - FAKE_WARNING Removing "C:/Users/appveyor/AppData/Roaming/Xpiks/Xpiks_integration-tests/HEAD/session-06122019-215843/plugins/xpiks-helloworld-plugin0.dll" failed, error: "Access is denied."
21:59:13.558 debug T#2604 Helpers::removeFile - Attempting to remove file "C:/Users/appveyor/AppData/Roaming/Xpiks/Xpiks_integration-tests/HEAD/session-06122019-215843/plugins/xpiks-helloworld-plugin0.xpiksplugin"
21:59:13.558 info T#2604 Plugins::PluginPackage::uninstall - FAKE_WARNING Package "C:/Users/appveyor/AppData/Roaming/Xpiks/Xpiks_integration-tests/HEAD/session-06122019-215843/plugins/xpiks-helloworld-plugin0.xpiksplugin" uninstalled with error, see previous logs
21:59:13.558 info T#2604 Plugins::PluginManager::removePlugin - FAKE_WARNING Failed to remove plugin "HelloWorldPlugin"
21:59:13.558 info T#2604 Plugins::PluginConfig::remove - Try to remove plugin: "C:/Users/appveyor/AppData/Roaming/Xpiks/Xpiks_integration-tests/HEAD/session-06122019-215843/plugins/xpiks-helloworld-plugin0.dll"
21:59:13.558 debug T#2604 Plugins::PluginConfig::remove - Removing plugin "C:/Users/appveyor/AppData/Roaming/Xpiks/Xpiks_integration-tests/HEAD/session-06122019-215843/plugins/xpiks-helloworld-plugin0.dll"
21:59:13.558 debug T#2604 Common::DelayedActionEntity::justChanged - started timer 2
21:59:13.558 info T#2604 Plugins::PluginManager::removePlugin - Plugin with ID # 0 removed
21:59:13.558 info T#2604 Plugins::PluginManager::removePlugin - Can unload the plugin: false
```

ribtoks:
I fixed issues on Windows with load/unload of dll and merged all your changes in #795 
Abandoning this PR.

PR #795, opened 2019-12-08
Velmafia dev/plugin json

ribtoks:


ribtoks:
@velmafia Overall you couldn't unload the plugin because in QPluginLoader [by default since Qt 5.7, QLibrary::PreventUnloadHint is set](https://doc.qt.io/qt-5/qpluginloader.html#loadHints-prop). So it was simply disallowing unloading. Before in Xpiks it was tolerated by ignoring the results.

Also - I fixed failing core tests by removing tests that were failing. On the serious note, core tests is not a place to test something with filesystem so it was not appropriate to test unzipping etc.

Overall I changed semantics of `PluginConfig` to store vector of `std::shared_ptr` and even though it is incorrect from the ownership perspective, as you noted, it fixed all other semantic issues and even the bug where you modified the vector during "for" loop (in `PluginManager::loadPlugins()`).

Please take a look at my latest changes if you see some potential for improvement or you have some comments. Otherwise I'm merging as soon as it's green.

PR #806, opened 2019-12-16
GitHub actions for linux and osx

velmafia:
Cache for `apt`/`brew` give nothing (maybe a few second), so I gave up on him, it will be better to use this [2GB cache space](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy) for `ccache`. 
Ccache make it really fast:
- `linux` full CI ~ 8-9 min;
- `osx` full CI ~ 15-18 min;

This is proposal, still WIP.


ribtoks:
If Linux and macOS build just works, can we merge it today? What is missing there? (e.g. codecov integration should be relatively trivial..)

.github/workflows/CI-linux.yaml
```diff
@@ -0,0 +1,42 @@
...
+on: [push]
+
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        target: [xpiks, core-tests, integration-tests, ui-tests]
+
+    runs-on: ubuntu-latest
```
ribtoks: `-latest` is probably not a way to have a repeatable build. Can we somehow point to specific release?

velmafia: This was copy-paste from some example, now is pined to specific release.

velmafia:
I mean not exactly something is missing, but more this is my first GitHub Actions, and my conceptions was changed a lot of time.

I fix `codecov` but now you need to [add token](https://docs.codecov.io/v4.3.0/docs/about-the-codecov-bash-uploader#section-upload-token) to `Settings => Secrets` (`CODECOV_TOKEN` variable).

Also I override `travis` variable (`TRAVIS_BUILD_DIR`, `TRAVIS_OS_NAME`) because I don't want change this in all script, I think this can be changed (by `sed` replace) when `travis-ci` will be disabled, or now I can rename this variable to something like `BUILD_DIR` and `OS_NAME`. What do you think?

And one more think:
Now I see that `ui-tests` failed for Linuks:
```
QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-runner'
qt.qpa.screen: QXcbConnection: Could not connect to display :109
Could not connect to any X display.
```
It's occurs (probably) after rebase, but I don't investigate it yet.

ribtoks:
I added codecov token, thanks.

Also I fixed ui tests on linux.

I think it makes sense to get rid of travis naming if we will be using GitHub Actions. But yeah, I understand it could be a pain. Maybe a separate PR? But it should be better done sooner rather than later since it's much worse to maintain 2 of them.

velmafia:
> I think it makes sense to get rid of travis naming if we will be using GitHub Actions. But yeah, I understand it could be a pain. Maybe a separate PR?

Yes, separate PR is a good idea. I can do this after porting `windows` to `action`. This could be general cleanup after `travis-ci`/`appveyor`. 

(But I propose to use parallel CI for some time, to check if `actions` working the same as previous CI's).

ribtoks:
Ok, no problem

ribtoks:
Do you know where can I see a link to green GitHub Action for this PR?

velmafia:
`Actions` probably have different access permission than whole repo.
I'm able to see my "actions" when I'm logged to GH and click `Actions` tab, but when I try to see it from Firefox private windows have 404 :/
Here is direct link to my last commit: https://github.com/velmafia/xpiks/commit/f22439db1008f2063e3b7c818f44fa7cdd173873/checks

PS.
> Also I fixed ui tests on linux.

I don't see this commit in PR, so probably this repo should be rebased to fix `ui-test` on Linux.

velmafia:
OK, I rebased this PR, and `ui-test` still failed: https://github.com/velmafia/xpiks/commit/fbfac1d190d24bab3ae61be43df40b743fb4cf20/checks
I will check this later.

ribtoks:
Yes, they failed. Probably the problem was different than I thought, sorry.

